PR #4, opened 2018-03-12
Fixed: user tokens not cleared on logout if `IdentityManager` is not available

lrossi:
Fixed by making the User object responsible for managing the tokens.


Example/Tests/IdentityManagerTests.swift
```diff
@@ -77,6 +77,18 @@ class IdentityManagerTests: QuickSpec {
...
+                Utils.createDummyKeychain()
+                let user = Utils.makeIdentityManager().currentUser
+                expect(user.state).to(equal(UserState.loggedIn))
+                user.logout()
+
+                let tokens = UserTokensKeychain().data().first
+                expect(tokens?.accessToken).to(beNil())
+                expect(tokens?.refreshToken).to(beNil())
+                expect(tokens?.idToken).to(beNil())
+            }
```
aliak00: ❤️

Example/Tests/UserTests.swift
```diff
@@ -33,13 +33,12 @@ class UserTests: QuickSpec {
...
 
-            it("Should emit set event when session refreshes it") {
+            it("Should update tokens when session refreshes it") {
                 self.stub(uri("/oauth/token"), try! Builders.load(file: "valid-refresh", status: 200))
                 let user = TestingUser(state: .loggedIn)
                 let oldTokens = user.wrapped.tokens
-                var newTokens: TokenData?
-                _ = user.wrapped.didSetTokens.register { newTokens = $0.new }
                 user.refresh { _ in }
+                let newTokens = UserTokensKeychain().data().first
```
aliak00: Should this be `let newTokens = user.wrapped.tokens` ?

Or maybe by "update tokens" in the test name it's meant that it should be updated in the keychain?

Example/Tests/UserTests.swift
```diff
@@ -177,43 +176,42 @@ class UserTests: QuickSpec {
...
+            it("Should update tokens if set differently") {
                 let user = User(state: .loggedOut)
-                var newTokens: TokenData?
                 _ = try? user.set(accessToken: "hehe", refreshToken: "hehe", idToken: "hehe", userID: "hehe")
-                _ = user.didSetTokens.register { newTokens = $0.new }
-
                 _ = try? user.set(accessToken: "hehe again")
-                expect(newTokens).toNot(beNil())
-                newTokens = nil
+                var tokens = UserTokensKeychain().data().first
```
aliak00: Should you get these from the user instead of see if they're updated in there?

Example/Tests/UserTests.swift
```diff
@@ -177,43 +176,42 @@ class UserTests: QuickSpec {
...
+            it("Should not change tokens if set to same") {
                 let user = User(state: .loggedOut)
-                var newTokens: TokenData?
                 _ = try? user.set(accessToken: "hehe", refreshToken: "hehe", idToken: "hehe")
-                _ = user.didSetTokens.register { newTokens = $0.new }
+                let tokens = UserTokensKeychain().data().first
+                expect(tokens).toNot(beNil())
                 _ = try? user.set(accessToken: "hehe", refreshToken: "hehe", idToken: "hehe")
-                expect(newTokens).to(beNil())
+                expect(UserTokensKeychain().data().first).to(equal(tokens))
```
aliak00: fyi: i think you can also do `expect(tokens) == tokens` instead of using `.to(equal...`

Example/Tests/UserTests.swift
```diff
@@ -177,43 +176,42 @@ class UserTests: QuickSpec {
...
-                var newTokens: TokenData?
                 _ = try? user.set(accessToken: "hehe", refreshToken: "hehe", idToken: "hehe", userID: "hehe")
-                _ = user.didSetTokens.register { newTokens = $0.new }
-
                 _ = try? user.set(accessToken: "hehe again")
-                expect(newTokens).toNot(beNil())
-                newTokens = nil
+                var tokens = UserTokensKeychain().data().first
+                expect(tokens).toNot(beNil())
+                UserTokensKeychain().removeTokens(forAccessToken: tokens?.accessToken ?? "")
```
aliak00: removeTokens I guess doesn't really do anything here since the tokens are memory based in a user object and only persisted and unpersisted on set/logout right? Should we maybe just use more of user.tokens instead of `UserTokensKeychain().data().first` to get the tokens that are in the user object instead?

Example/Tests/UserTests.swift
```diff
@@ -215,6 +215,50 @@ class UserTests: QuickSpec {
...
+            }
+
+            it("Should update keychain when user is refreshed") {
+                self.stub(uri("/oauth/token"), try! Builders.load(file: "valid-refresh", status: 200))
+                var newTokens: TokenData?
+                do {
+                    Utils.createDummyKeychain()
+                    let user = User(state: .loggedOut)
+                    try? user.loadStoredTokens()
+                    user.refresh { _ in }
```
aliak00: `User.refresh` is asynchronous so I guess this could potentially fail `TestingUser` implements the same methods as `User` but synchronously instead.

aliak00:
I thought maybe renaming the PR could get a commit message that is by the guidelines, but apparently not, so can maybe change a relevant commit to reflect the commit guidelines?

"Fixed: blah blah bug where blah blah" or something

lrossi:
Hmm, what if the actual fix is spread across multiple commits? 😛

lrossi:
Replaced by #11

PR #6, opened 2018-03-13
generic error value is not really optional, so don't declare it optional

aliak00:


SchibstedAccount/Core/GenericError.swift
```diff
@@ -20,16 +20,13 @@ extension GenericError {
 
 extension GenericError.Unexpected: CustomStringConvertible {
     var description: String {
-        if let string = self.string {
-            return "\(type(of: self))! \(string)"
-        }
-        return "\(type(of: self))!"
+        return "\(type(of: self))! \(string)"
```
lrossi: Even though it should never fail, can we avoid the bang here? Maybe something like `?? "n/a"`?

SchibstedAccount/Core/GenericError.swift
```diff
@@ -38,9 +35,6 @@ extension GenericError {
 
 extension GenericError.WTF: CustomStringConvertible {
     var description: String {
-        if let string = self.string {
-            return "\(type(of: self))! \(string)"
-        }
-        return "\(type(of: self))!"
+        return "\(type(of: self))! \(string)"
```
lrossi: Same here

SchibstedAccount/Core/GenericError.swift
aliak00: The bang is part of the output text :p 

i.e. "Error! something failed"

lrossi: Oh, true! 😆

PR #7, opened 2018-03-13
Add ability to choose whether user's login should be persistent

lrossi:
This PR adds the option to the low-level calls and provides a simple UI for testing the functionality in the Example app.

A separate PR will then expose the functionality in the SDK's UI.


SchibstedAccount/Manager/User/User.swift
```diff
@@ -49,6 +49,8 @@ public class User: UserProtocol {
     ///
     public weak var delegate: UserDelegate?
 
+    var isPersistent = false
```
aliak00: should this do stuff on set? i.e.:

```
didSet {
  if newValue { storeTokens } else { clearTokens }
}

?

lrossi: After merging the other PR that makes `User` responsible of managing the storage of the tokens, `isPersistent` is now a private state

aliak00: sweet! Makes more sense

PR #11, opened 2018-03-13
Fix user tokens not cleared on logout if `IdentityManager` is not available

lrossi:
Take # 2 🎬

### DIFF
--- a/Example/Tests/UserTests.swift
+++ b/Example/Tests/UserTests.swift
@@ -215,6 +215,50 @@ class UserTests: QuickSpec {
             }
         }
 
+        describe("Keychain management") {
+            it("Should not treat the user as logged in if it does not find keychain data") {
+                let user = User(state: .loggedOut)
+                expect(user.state).to(equal(UserState.loggedOut))
+            }
+
+            it("Should treat the user as logged in if it finds keychain data") {
+                Utils.createDummyKeychain()
+                let user = User(state: .loggedOut)
+                try? user.loadStoredTokens()
+                expect(user.state).to(equal(UserState.loggedIn))
+            }
+
+            it("Should update keychain when user is refreshed") {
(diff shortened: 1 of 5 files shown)


Example/Tests/UserTests.swift
```diff
@@ -177,43 +176,86 @@ class UserTests: QuickSpec {
...
+                expect(user.tokens).toNot(beNil())
+                expect(user.tokens?.userID) == "hehe again"
+            }
+        }
 
-                _ = try? user.set(userID: "hehe again")
-                expect(newTokens).toNot(beNil())
+        describe("Keychain management") {
+            it("Should not treat the user as logged in if it does not find keychain data") {
+                let user = User(state: .loggedOut)
```
aliak00: If you specify state loggedOut then User will not even try and set tokens. I guess you're missing a call to `try? user.loadStoredTokens()`?

Example/Tests/UserTests.swift
```diff
@@ -177,43 +176,86 @@ class UserTests: QuickSpec {
...
+                    let user = TestingUser(state: .loggedOut)
+                    try? user.wrapped.loadStoredTokens()
+                    user.refresh { _ in }
+                    newTokens = user.wrapped.tokens
+                }
+
+                let tokens = UserTokensKeychain().data().first
+                expect(newTokens?.accessToken).to(equal(tokens?.accessToken))
+                expect(newTokens?.refreshToken).to(equal(tokens?.refreshToken))
+                expect(newTokens?.idToken).to(equal(tokens?.idToken))
```
aliak00: You can make this test more solid by getting the keychain data before the update as well and making sure they are not the equal.

Example/Tests/UserTests.swift
lrossi: Correct 👍

PR #17, opened 2018-03-19
Standalone project without cocoapods

Torsph:
Makes Carthage work without exposing the Pods.xcodeproj created by CocoaPods.

Tests should be moved into the main project when we remove the Mockingjay dependency. Mockingjay does not support Carthage today and the project owner will not support it.


aliak00:
Will need path changes of `SchibstedAccount/` -> `Sources/` in:
- scripts/format.sh
- scripts/git-hooks/pre-commit
- codecov.yml
- swiftlint.yml

PR #19, opened 2018-03-20
Logo

aliak00:


SchibstedAccount/UI/Views/LogoView.swift
```diff
@@ -0,0 +1,39 @@
...
+        let filler = UIView(frame: self.frame)
+        filler.translatesAutoresizingMaskIntoConstraints = false
+        filler.widthAnchor.constraint(equalToConstant: self.frame.width).isActive = true
+
+        if let titleLogo = theme.titleLogo {
+            let customLogoImageView = UIImageView(image: titleLogo)
+            let scale = titleLogo.size.width / titleLogo.size.height
+            customLogoImageView.contentMode = .scaleAspectFit
+            customLogoImageView.heightAnchor.constraint(equalToConstant: 20).isActive = true
+            customLogoImageView.widthAnchor.constraint(equalToConstant: 20 * scale).isActive = true
```
lrossi: Fractions of pixels are tricky, safer to `ceil()`

aliak00:
Changed the height back to what it was, so should be back to normal now! 

Nice catch!

PR #25, opened 2018-03-22
Handle updated terms

lrossi:
Handle update in terms & conditions by presenting an acceptance screen at app's startup for logged-in users.

- ~New methods are provided in the `User` class to fetch acceptance status and accept new terms (the latter only intended for the headless approach).~
- New ~UI route~ `IdentityUI`'s method is provided to present the "accept terms" screen. Logout is silently performed in case the user closes the modal screen before accepting terms (we need new guidelines on this).
- ~`IdentityManager` is no more passed in the initializer of `IdentityUI`, since we need to make sure the route is passed the exact same user the app has obtained from its own instance of `IdentityManager`, otherwise they won't get logout notifications.~
- ~A new loading screen is displayed while the updated terms are being fetched.~ Client is responsible for fetching the updated terms and hand them over to the SDK.
- ~`trackerViewID` is now optional in `IdentityUIViewController` since the loading screen is not tracked.~


SchibstedAccount/Manager/User/User.swift
```diff
@@ -124,6 +134,40 @@ public class User: UserProtocol {
...
+     case the user is logged out for not having accepted the new terms.
+
+     If you are using the headless approach instead, you should then present your own UI and manually call `acceptNewTerms(:)`, if the user accepted the new
+     terms, or `logout()`, if the user rejected them.
+
+     - parameter completion: Callback that is called after acceptance status has been checked.
+     */
+    public func hasAcceptedLatestTerms(completion: @escaping BoolResultCallback) {
+        self.agreements.status(completion: completion)
+    }
```
aliak00: Yeah, maybe just add comments to `AgreementsAPI.status` instead of a wrapper function.

SchibstedAccount/Manager/User/User.swift
```diff
@@ -124,6 +134,40 @@ public class User: UserProtocol {
...
+    /**
+     Tells the server that the user has accepted the latest terms and conditions.
+
+     If you use the recommended way of presenting the screen to accept the terms, i.e. if you the UI flows provided by `IdentityUI`, then you should *not* call
+     this method manually, as it will be automatically called for you when the user accepts the new terms. You should only use this method if you are following
+     the headless approach and implementing your own UI.
+
+     - parameter completion: Callback that is called after the acceptance of the terms has been sent to the server.
+     */
+    public func acceptNewTerms(completion: @escaping NoValueCallback) {
```
aliak00: Same here.

SchibstedAccount/UI/Coordinators/UpdatedTermsCoordinator.swift
```diff
@@ -0,0 +1,124 @@
+//
+// Copyright 2011 - 2018 Schibsted Products & Technology AS.
+// Licensed under the terms of the MIT license. See LICENSE in the project root.
+//
+
+import UIKit
+
+class UpdatedTermsCoordinator: FlowCoordinator {
```
aliak00: Can we not use the existing CompleteProfileCoordinator? - it will show the screen and then it just won't do any of the required fields entry since there won't be any (if there are then it's also probably the correct thing to do with gdpr, etc, anyway).

SchibstedAccount/UI/IdentityUIRoute.swift
```diff
@@ -19,10 +19,16 @@ extension IdentityUI {
         /// Route to the next step after email validation occurred after signup.
         case validateAuthCode(code: String, shouldPersistUser: Bool)
 
+        /// Route to present updated terms and condition to an already logged-in user.
+        case presentUpdatedTerms
```
aliak00: Should this just be `presentTerms` maybe? There should be no difference from the SDK point of view of updated terms or terms. They are also pretty much the same data/screen in the backend. And this should ideally have an associated property of `User`.

I realize that our current design is not going to work with only passing User here, so it should probably be IdentityManager goes in here :(

SchibstedAccount/UI/IdentityUIViewController.swift
```diff
@@ -66,7 +66,7 @@ class IdentityUIViewController: UIViewController {
         return ""
     }
 
-    let trackerViewID: TrackingEvent.View
+    let trackerViewID: TrackingEvent.View?
```
aliak00: Why'd this change to optional?

aliak00:
Seems like a bit of a complicated task :p Biggest thing here would be:
* Can we keep IdentityManager in initializer?
* How can we use the complete profile flow, seems like that's what we want do do
* The route to complete profile I guess has to contain a User object ideally, but it seems we have to give it an IdentityManager object instead. Can this be avoided though?

DemoApp/SchibstedAccountDemo/AppDelegate.swift
```diff
@@ -110,10 +110,44 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+                // Present UI to accept new terms.
+                guard let vc = self?.window?.rootViewController else {
+                    return
+                }
+                self?.identityUI = IdentityUI(configuration: .default)
+                self?.identityUI?.delegate = self
+                
+                // It is crucial that you pass the same instance of `IdentityManager` used to obtain the current logged-in user (that is stored in `self.user`
+                // in this example app), otherwise you won't get logout notifications for that user in case the user is logged out for not having accepted the
+                // new terms.
```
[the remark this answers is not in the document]
lrossi: Not sure what the ideal thing would be. In a way, I would kind of expect to get back the same instance of `User` if I am fetching the same user and the previous instance is still retained by someone else (a little bit like how `NSManagedObjectContext` works with `NSManagedObject`-s). That being the case, the actual `IdentityManager` you use wouldn't be so important.

But we can think this more thoroughly, in the meantime storing `identityManager` sounds like a good compromise.

Example/SchibstedAccount/AppDelegate.swift
```diff
@@ -158,7 +158,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         }
 
         if let vc = self.window?.rootViewController, let identityUI = self.statusViewController?.identityUI, let route = IdentityUI.Route(payload: payload) {
-            identityUI.presentIdentityProcess(from: vc, route: route)
+            identityUI.presentIdentityProcess(from: vc, route: route, identityManager: nil)
```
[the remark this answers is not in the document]
lrossi: Alternatively, I think, we could require an `IdentityManager` to be passed to the initializer (thus removing the init variant without the manager).

SchibstedAccount/Manager/AppLaunchData.swift
```diff
@@ -99,4 +99,14 @@ extension AppLaunchData {
...
         return nil
     }
+
+    /**
+     Returns whether the given launch options contain a URL that is recognized as a deep link to the SchibstedAccount SDK.
+
+     - parameter launchOptions: The launch options you get through `UIApplicationDelegate.application(_:url:options:)`.
+     - parameter clientConfiguration: The current client configuration.
+     */
+    public static func doesLaunchOptionsContainRecognizedURL(_ launchOptions: [AnyHashable: Any]?, for clientConfiguration: ClientConfiguration) -> Bool {
```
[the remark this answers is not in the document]
lrossi: Yeah, it's exactly the same. I added it only because after writing `if let launchData = AppLaunchData(launchOptions)` I didn't find its meaning so obvious 😛

SchibstedAccount/Manager/User/User.swift
lrossi: Yes, no need for the wrapper 👌

SchibstedAccount/UI/Coordinators/UpdatedTermsCoordinator.swift
lrossi: I was thinking that ideally `CompleteProfileCoordinator` should have used its own instance of `UpdatedTermsCoordinator` in a subsequent refactoring, but if it's a good thing to check for the required fields as well then probably `CompleteProfileCoordinator` can be adapted to work in this case as well.

SchibstedAccount/UI/IdentityUIRoute.swift
lrossi: Ok for `presentTerms`.

Yeah, so I was passing the `IdentityManager` to the `presentRoute...` method instead. But maybe it can now go back to the initializer.

SchibstedAccount/UI/IdentityUIViewController.swift
lrossi: Cause the new `Loading` screen doesn't have a tracker ID.

DemoApp/SchibstedAccountDemo/AppDelegate.swift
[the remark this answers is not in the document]
aliak00: Aye, maybe we can extract a user from User.globalStore!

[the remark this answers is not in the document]
aliak00: Wait, maybe not...

SchibstedAccount/UI/Coordinators/UpdatedTermsCoordinator.swift
aliak00: Ahh, yeah ok. I'm not sure if it's a good thing anymore actually :p

SchibstedAccount/UI/IdentityUIViewController.swift
aliak00: ahh

DemoApp/SchibstedAccountDemo/AppDelegate.swift
```diff
@@ -132,18 +134,56 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+                self?.identityManager.fetchTerms { [weak self] result in
+                    switch result {
+                    case let .success(terms):
+                        // Present UI to accept new terms.
+                        guard let viewController = self?.window?.rootViewController, let user = self?.identityManager.currentUser else {
+                            return
+                        }
+                        // It is important that you pass the same instance of `User` that you previously stored, otherwise you won't get logout notifications
+                        // for that user in case the user is logged out for not having accepted the new terms.
+                        IdentityUI.presentTerms(terms, for: user, from: viewController, configuration: .current)
```
aliak00: User has an internal configuration object as well, so it's probably better to just use that one?

Source/UI/IdentityUI.swift
```diff
@@ -105,6 +105,51 @@ public class IdentityUI {
...
+     - parameter: user: The user that requires acceptance of new terms. It is important that you pass the same instance of `User` you previously obtained from
+       an `IdentityManager` and stored, otherwise you won't get logout notifications for that user in case the user is logged out for not having accepted the
+       new terms.
+     - parameter: viewController: The view controller to present the screen from.
+     - parameter: configuration: The UI configuration.
+     */
+    public static func presentTerms(_ terms: Terms, for user: User, from viewController: UIViewController, configuration: IdentityUIConfiguration) {
+        guard self.presentedIdentityUI == nil, self.updatedTermsCoordinator == nil else {
+            // Another screen of the Identity UI is already presented.
+            return
```
aliak00: Should we puke some stuff here since this should in theory never happen? precondifitionFailure maybe or just a print statement or something?

DemoApp/SchibstedAccountDemo/AppDelegate.swift
aliak00: Ah, it's a UI config, not a client config :) 

Should this be an error if `user.clientCOnfiguration != configuration.clientConfiguration` => `.failure(something)` ?

lrossi: Oh, good point! Ideally, I would really prefer enforcing a coherent state through the APIs rather than trowing an error if the client makes a mistake, but it doesn't look trivial in this case.

Btw, is there a specific reason why the client configuration is owned by the UI configuration, or could they be treated more independently? Either way that is of course out of scope, so I guess an error has to be ok for now.

Source/UI/IdentityUI.swift
lrossi: Very theoretically, the user could log out and start a new login flow while the network calls to fetch the acceptance status and the terms are made, so in that particular case I think it might be possible you end up here with another flow already in place and you just want to ignore this one. But I should probably check and fail if the user is still logged in && another flow is in place, as that should surely never happen.

DemoApp/SchibstedAccountDemo/AppDelegate.swift
aliak00: No particular reason other than it felt more friendly to have client pass one configuration object instead of two.

lrossi:
@aliak00 take a look at the newest commits when you have a chance 🙂

PR #33, opened 2018-03-26
Fix .gitignore

Torsph:


Torsph:
Yes! @aliak00 will do it after easter 🐣

PR #43, opened 2018-04-06
Added: convenience initializer for copy construction for IdentityUICo…

aliak00:
…nfiguration

This will allow apps to take an existing configuration and use that as
default values in cases where they want to alter a single property but
want to keep the rest of the settings the same


aliak00:
Agreed, done.

Source/UI/IdentityUIConfiguration.swift
```diff
@@ -48,18 +48,50 @@ public struct IdentityUIConfiguration {
...
         self.tracker = tracker
+        if let appName = appName {
+            self.appName = appName
+        }
+    }
+
+    /**
+     Call this to replace "parts" of the configuration
+
+     - parameter copyOf: the `IdenitytUIConfiguration` you want to start off with as a copy, you can then supply the values you want to override
```
lrossi: Extra parameter 😛

aliak00: fixed. Thanks!

Issue #47, opened 2018-04-11
Set modalPresentationStyle for iPad


eriksundin:
Currently we're only able to present the UI in full-screen mode which looks so-so on iPad.
Feature request, expose a way to set the modal presentation style of the UI.

In our specific case, so we can achieve:
`modalPresentationStyle = .formSheet`

Cheers

aliak00:
Nice. I think these are good ideas! Especially the hook. Fullscreen UI on an iPad is weird indeed.

PR #52, opened 2018-04-17
ADDED: presentationHook block to IdentityUIConfiguration

eriksundin:
Proposed solution for #47


aliak00:
Noice! You'll also have to add it here I believe:

https://github.com/eriksundin/account-sdk-ios/blob/27b9f2ab71b64094e5a007fa8ddf5068a5048ad8/Source/UI/IdentityUI.swift#L151

Then I think this is good to go!

eriksundin:
@aliak00 Cool, thanks. Added.

aliak00:
🎉

PR #62, opened 2018-04-23
Info screen

aliak00:


aliak00:
Temporary until we get asset from design!

PR #63, opened 2018-04-23
Merchant api

aliak00:


Source/Manager/IdentityAPI/IdentityAPI.swift
```diff
@@ -343,6 +343,8 @@ class IdentityAPI {
             return .invalidPhoneNumber
         case let .object("ApiException", .string(string), 302):
             return .alreadyRegistered(message: string)
+        case let .object("ApiException", .string(string), 404) where string == "No access to that product.":
```
lrossi: Does 404 mean no access? 🤔

aliak00: Technically means not found, but yeah, for a product that you don't have access to this is what spid does (it seems).

Issue #56, opened 2018-04-23
Show a screen about changes in data privacy


mela124:
Design:
https://app.zeplin.io/project/5ad8456d755b8f2e0da01dcc/screen/5ad84fe09e46a4341c0a257e
Full text: https://app.zeplin.io/project/5ad8456d755b8f2e0da01dcc/screen/5ad85a279e46a4341c0a92c0

Note: Disregard header, no changes on iOS header.

Acceptance criteria:

-It should only be shown once to the user (we can also do it once by platform. ie they can see it on all platforms they log in to)
-It is always shown before the terms of use page (sign up, login and no terms has been accept yet, updated terms) - it's a combo with terms of use page
-Continue button is enabled by default
-Continue button is sticky at the bottom of the screen. User can scroll through the text
- On press of Continue, go to the terms screen
-Back and exit controls should be consistent behaviour as rest the of the screens. In case of updated terms, this is considered as the 1st screen so just the "X" button

Note: disregard the link at the end of the screen first, we will create separate task to add all links once they are ready

lrossi:
@mela124 In case of updated terms, does the automatic logout apply to this screen as well if the user closes it? If so, do we provide some kind of message?

mela124:
Hello! Good question, no, if they close this page they will still be logged in but the terms page will show up next if there’s terms to be accepted.

lrossi:
Do you mean an additional modal screen with just the terms should show up straight away in case this one is closed? Or is it ok to show the terms next time the app is launched?

Or maybe, if this is just an info alert with no action involved, we could avoid canceling and only have a `Continue` button so that the terms screen is always pushed next, and then that screen has a cancel button. What do you think?

mela124:
We will no longer need this screen. There is no need for do this task anymore. I will close.

PR #65, opened 2018-04-24
Refactor: add common ShowTermsCoordinator

lrossi:
Add a `ShowTermsCoordinator` that is used by both `CompleteProfileCoordinator` and `UpdatedTermsCoordinator`. This was originally intended for #56, but might be useful regardless, as it should make changing the terms screen simpler by providing it from a single place.

### DIFF
--- a/Source/UI/Coordinators/CompleteProfileCoordinator.swift
+++ b/Source/UI/Coordinators/CompleteProfileCoordinator.swift
@@ -168,52 +168,33 @@ extension CompleteProfileCoordinator {
 
             switch result {
             case let .success(terms):
-                self?.showAcceptTermsView(for: terms, on: loginFlowVariant, completion: completion)
+                self?.spawnShowTermsCoordinator(for: terms, on: loginFlowVariant, completion: completion)
             case let .failure(error):
                 if self?.handle(error: error) != true { completion(.error(error)) }
             }
         }
     }
 
-    private func showAcceptTermsView(
+    private func spawnShowTermsCoordinator(
         for terms: Terms,
         on loginFlowVariant: LoginMethod.FlowVariant,
         completion: @escaping (AgreementsAcceptanceResult) -> Void
     ) {
-        let navigationSettings = NavigationSettings(
-            cancel: configuration.isCancelable ? { completion(.cancel) } : nil,
-            back: { completion(.back) }
-        )
-        let viewModel = TermsViewModel(
-            terms: terms,
-            loginFlowVariant: loginFlowVariant,
-            appName: self.configuration.appName,
-            localizationBundle: self.configuration.localizationBundle
-        )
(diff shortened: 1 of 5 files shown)


Source/UI/Coordinators/CompleteProfileCoordinator.swift
```diff
@@ -168,52 +168,33 @@ extension CompleteProfileCoordinator {
...
-        )
+        let showTermsCoordinator = ShowTermsCoordinator(navigationController: self.navigationController, configuration: self.configuration)
+        let input = ShowTermsCoordinator.Input(terms: terms, loginFlowVariant: loginFlowVariant, presentationStyle: .pushToExistingNavigationFlow)
 
-        let viewController = TermsViewController(configuration: self.configuration, navigationSettings: navigationSettings, viewModel: viewModel)
-        viewController.didRequestAction = { [weak self] action in
-            switch action {
-            case .acceptTerms:
+        self.child = ChildFlowCoordinator(showTermsCoordinator, input: input) { [weak self] output in
+            self?.child = nil
```
aliak00: Man, if you hadn't added this line I would've never remembered about having to do it 😱

Source/UI/Coordinators/ShowTermsCoordinator.swift
```diff
@@ -0,0 +1,102 @@
...
+            case .cancel:
+                completion(.cancel)
+            }
+        }
+
+        switch presentationStyle {
+        case .pushToExistingNavigationFlow:
+            self.navigationController.pushViewController(viewController, animated: true)
+        case .replaceNavigationFlow:
+            self.navigationController.viewControllers = [viewController]
```
aliak00: Do we need the presentation style? Is it enough to base this on whether `self.navigationController.viewControllers.count > 0` ? If yes then we continue a flow, else we're starting a new flow.

aliak00: Existence of back button is also then implicit.

aliak00: Or now I'm thinking maybe there's something with a loginFlowVariant we're missing. Because there's no variant on the "static present updated terms flow" ?

lrossi: Yes that check should work, I'll change it 👌

True, the `loginFlowVariant` doesn't really apply when presenting the updated terms, for now it falls back to the login variant in that case.

Source/UI/Coordinators/CompleteProfileCoordinator.swift
lrossi: Yeah, that's scary... I think I just had an idea, I'll try if that works (in a separate PR).

PR #67, opened 2018-04-24
Update copy in required fields screen

### LINKED ISSUE #59: Add copy on Required Fields screen
Design: https://app.zeplin.io/project/5ad8456d755b8f2e0da01dcc/screen/5ad84fdef27084a923eaa419
Note: Disregard header, no changes in the header for iOS

Acceptance criteria:
- Add copy on the required fields screen

Note: 
- Do not add the links yet (but show the copy). We will create a separate ticket for all links.
- Do not add more required fields for now

lrossi:
Use a text view instead of a label since it will need to handle links.

Closes #59.

### DIFF
--- a/Source/UI/Screens/RequiredFields/RequiredFieldsViewController.swift
+++ b/Source/UI/Screens/RequiredFields/RequiredFieldsViewController.swift
@@ -21,7 +21,6 @@ class RequiredFieldsViewController: IdentityUIViewController {
     @IBOutlet var subtext: TextView! {
         didSet {
             self.subtext.isEditable = false
-            self.subtext.delegate = self
             self.subtext.attributedText = NSAttributedString(string: self.viewModel.subtext)
         }
     }
@@ -231,9 +230,6 @@ class RequiredFieldsViewController: IdentityUIViewController {
         super.endLoading()
         self.continueButton.isAnimating = false
     }
-}
-
-extension RequiredFieldsViewController: UITextViewDelegate {
 }
 
 extension RequiredFieldsViewController: UITextFieldDelegate {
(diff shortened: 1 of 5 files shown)


Source/UI/Screens/RequiredFields/RequiredFieldsViewController.swift
```diff
@@ -231,6 +233,9 @@ class RequiredFieldsViewController: IdentityUIViewController {
     }
 }
 
+extension RequiredFieldsViewController: UITextViewDelegate {
```
aliak00: forgot to delete extension or forgot to add members :p

Source/UI/Views/TextView.swift
```diff
@@ -4,14 +4,19 @@
 //
 
 class TextView: UITextView, Themeable {
```
aliak00: Can this TextView also be used for the terms screen (since they show links as well)?

aliak00: Nevermind, didn't realize this was there already 🤦‍♂️

Source/UI/Screens/RequiredFields/RequiredFieldsViewController.swift
lrossi: I added it as a placeholder since it'll be needed to handle links, and by having it I could already set the view controller as the text view's delegate. Too far maybe? 😛

aliak00: Haha maybe a bit too eager 😂 I'll leave this to you to merge and add to in a follow up PR then :)

PR #79, opened 2018-05-07
dont limit login password to 8 chars

### LINKED ISSUE #74: Remove password restriction for log in only
Situation: We currently still have users with less than 8 characters and currently they can't log in.
Expected results: Users can't log in with however length of password. Sign up with 8 characters still applies.

aliak00:
Fixes #74

### DIFF
--- a/Source/UI/Screens/Password/PasswordViewController.swift
+++ b/Source/UI/Screens/Password/PasswordViewController.swift
@@ -150,7 +150,7 @@ class PasswordViewController: IdentityUIViewController {
     }
 
     @IBAction func didClickContinue(_: Any) {
-        guard let password = self.password.text, password.count >= 8 else {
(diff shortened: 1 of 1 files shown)


Source/UI/Screens/Password/PasswordViewController.swift
```diff
@@ -150,7 +150,7 @@ class PasswordViewController: IdentityUIViewController {
     }
 
     @IBAction func didClickContinue(_: Any) {
-        guard let password = self.password.text, password.count >= 8 else {
+        guard let password = self.password.text, ((self.viewModel.loginFlowVariant == .signin && password.count >= 1) || password.count >= 8) else {
             self.showInlineError(.invalidUserCredentials(message: nil))
```
minarikg: Suggestion:         guard let password = self.password.text, ((self.viewModel.loginFlowVariant == .signin && password.count >= 1) || password.count >= 8) else {

PR #80, opened 2018-05-07
1.0.0 rc3 hot fix

aliak00:


Source/UI/Screens/Password/PasswordViewController.swift
```diff
@@ -137,7 +137,7 @@ class PasswordViewController: IdentityUIViewController {
     }
 
     @IBAction func didClickContinue(_: Any) {
-        guard let password = self.password.text, password.count >= 8 else {
+        guard let password = self.password.text, ((self.viewModel.loginFlowVariant == .signin && password.count >= 1) || password.count >= 8) else {
```
fespinoza: nitpicking: it's a bit difficult to understand what you are trying to achieve here, too complex logic for a guard clause

Source/UI/IdentityUI.swift
```diff
@@ -103,7 +103,7 @@ public class IdentityUI {
     // Used to store the currently presented identity process so that:
     // 1. The presentation of one process at a time can be enforced.
     // 2. When handling a universal link, the currently presented process (if any) can be retrieved.
-    private static var presentedIdentityUI: IdentityUI?
+    private weak static var presentedIdentityUI: IdentityUI?
```
fespinoza: it would be better that in the commit mention *why* you need this to be weak, to understand the context, what are you trying to solve explicitly?

fespinoza:
and you do have some merge conflicts to fix :P

Source/UI/Screens/Password/PasswordViewController.swift
aliak00: Can't argue with that :)

Source/UI/IdentityUI.swift
aliak00: True! I got lazy because this was a hot fix. The next release will have [this](https://github.com/schibsted/account-sdk-ios/pull/77) which does have a better commit message

aliak00:
Thanks fespinoza!

aliak00:
Marked branch as protected, tagged and released.

PR #105, opened 2018-05-16
Fixed: updated terms regression

aliak00:
Made it not weak and added a dismiss function incase the client needs to dismiss it prematurely.


aliak00:
Use #107 instead

PR #109, opened 2018-05-18
Fixed: Persist user tokens only after login process is complete

### LINKED ISSUE #102: Cancelling the terms screen keeps user logged in.
When implementing the SDK in my project I get an issue using the login UI. After the password is entered and the continue button is pressed I see the terms screen. When tapping the cancel button (the X in the top right corner) the callback claims the login was cancelled. However, when checking the  `identityManager.currentUser.state` the user seems to be logged in.

aliak00:
Closes #102

### DIFF
--- a/Source/Manager/User/User.swift
+++ b/Source/Manager/User/User.swift
@@ -156,7 +156,7 @@ public class User: UserProtocol {
             return
         }
 
-        self.updateStoredTokens(nil, previousTokens: oldTokens)
+        try? UserTokensStorage().clear(oldTokens)
         self.delegate?.user(self, didChangeStateTo: .loggedOut)
 
         self.api.logout(oauthToken: oldTokens.accessToken) { [weak self] result in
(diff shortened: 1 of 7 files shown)


Source/Manager/User/User.swift
```diff
@@ -225,15 +225,22 @@ public class User: UserProtocol {
 
         log(from: self, "new tokens \(newTokens)")
 
-        if let makePersistent = makePersistent {
-            self.isPersistent = makePersistent
+        self.isPersistent = makePersistent ?? self.isPersistent
+
+        let isNewLoggedInUser: Bool
+        if let newAnyUserID = newTokens.anyUserID, newAnyUserID != tokens.old?.anyUserID {
```
lrossi: Would `let isNewLoggedInUser = newTokens.anyUserID != nil && newTokens.anyUserID != tokens.old?.anyUserID` be clear enough?

Source/Manager/User/User.swift
```diff
@@ -251,17 +258,23 @@ public class User: UserProtocol {
         self.isPersistent = true
     }
 
-    func updateStoredTokens(_ tokens: TokenData?, previousTokens: TokenData?) {
-        // Always clear old tokens
+    private static func updatePersistedTokens(_ tokens: TokenData?, previousTokens: TokenData?) {
```
lrossi: Since the two operations are now decoupled, could it be clearer to separate this into two methods? (One for storing tokens and another one for clearing).

Source/UI/Interactors/OneTimeCodeInteractor.swift
```diff
@@ -18,8 +18,8 @@ class OneTimeCodeInteractor {
         self.identityManager.resendCode(to: identifier, completion: completion)
     }
 
-    func validate(oneTimeCode: String, for identifier: Identifier, persistUser: Bool, completion: @escaping (Result<User, ClientError>) -> Void) {
-        self.identityManager.validate(oneTimeCode: oneTimeCode, for: identifier, persistUser: persistUser) { [weak self] result in
+    func validate(oneTimeCode: String, for identifier: Identifier, completion: @escaping (Result<User, ClientError>) -> Void) {
+        self.identityManager.validate(oneTimeCode: oneTimeCode, for: identifier, persistUser: false) { [weak self] result in
```
lrossi: Should the `persistUser` flag be removed from `IdentityManager`'s method?

aliak00: Maybe if we make User.persistCurrentTokens public then? Can think of it in a different PR

PR #134, opened 2018-06-04
Fix travis test

aliak00:


Source/Core/Logger/Logger.swift
```diff
@@ -0,0 +1,302 @@
...
+
+ Two methods exist to allow for filtering of the log stream.
+
+ - `Logger.filterUnless(tag:)`
+ - `Logger.filterIf(tag:)`
+
+ */
+public class Logger {
+    /// Shared logger object
+    public static let shared: Logger = {
```
Torsph: You could just do `public static let shared = Logger()` if you don't need any extra setup.

Source/Core/Logger/Logger.swift
```diff
@@ -0,0 +1,302 @@
...
+    /// Removes all transports
+    public func removeTransports() {
+        self.queue.async { [weak self] in
+            self?.transports.removeAll()
+        }
+    }
+
+    /// Filters log messages unless they are tagged with `tag`
+    public func filterUnless(tag: String) {
+        _ = self.queue.async { [weak self] in
```
Torsph: Some places you do `_ = self.queue.async()`other places you don't. Maybe do the same all places?

Source/Core/Logger/Logger.swift
```diff
@@ -0,0 +1,302 @@
...
+        #endif
+
+        let thread = Thread.isMainThread ? "UI" : "BG"
+        let threadID = pthread_mach_thread_np(pthread_self())
+        let timestamp = self.startTime.elapsed
+        let string = "\(object())"
+
+        self.dispatchGroup.enter()
+        self.queue.async { [weak self] in
+            guard let strongSelf = self else { return }
```
Torsph: What happens to dispatchGroup when you call `enter()` and self is destroyed? Does it care, like holding on to something so it won't get destroyed?

Source/Core/Logger/Logger.swift
```diff
@@ -0,0 +1,302 @@
...
+            return value
+        }()
+
+        var allTags = [functionName, thread, fileName, level.rawValue]
+        allTags.append(contentsOf: explicitTags)
+        if let context = context {
+            allTags.append(context)
+        }
+
+        var shouldOutputToTransports = true
```
Torsph: Could both of these if checks be combined into one? And maybe just get assign and then make it a let instead of a var.

Source/Manager/User/User.swift
```diff
@@ -275,7 +275,7 @@ public class User: UserProtocol {
         }
 
         guard let refreshToken = tokens.refreshToken else {
-            forceLog(from: self, "no refresh token")
+            log(from: self, "no refresh token", force: true)
```
Torsph: Should this have level also?

Source/Manager/User/User.swift
```diff
@@ -264,7 +264,7 @@ public class User: UserProtocol {
             try UserTokensStorage().store(tokens)
             self.isPersistent = true
         } catch {
-            log(self, "failed to persist tokens: \(error)")
+            log(from: self, "failed to persist tokens: \(error)")
```
Torsph: Should this have level also?

Source/Manager/UserTokensStorage.swift
```diff
@@ -26,7 +26,7 @@ struct UserTokensStorage {
                 try self.store(tokens)
                 SPiDKeychainWrapper.removeAccessTokenFromKeychain(forIdentifier: kSPiDAccessToken)
             } catch {
-                forceLog(from: self, "failed to migrate \(tokens)")
+                log(from: self, "failed to migrate \(tokens)", force: true)
```
Torsph: Should this have level also?

Source/UI/IdentityUI.swift
```diff
@@ -185,7 +185,7 @@ public class IdentityUI {
 
             let msg = "Attempt to present updated terms while another Identity UI flow is already presented."
             assertionFailure(msg)
-            log(self, msg)
+            log(from: self, msg)
```
Torsph: Should this have level also?

Source/UI/IdentityUI.swift
```diff
@@ -353,7 +353,7 @@ extension IdentityUI: FlowCoordinator {
             guard case let .byRoute(route, _) = input else {
                 let msg = "Attempt to present a new Identity UI instance while another one is already presented."
                 assertionFailure(msg)
-                log(self, msg)
+                log(from: self, msg)
```
Torsph: Should this have level also?

Source/Manager/TaskManager/TaskManager.swift
```diff
@@ -54,8 +61,11 @@ class TaskManager {
                     return
                 }
 
+                log(level: .verbose, from: self, "did execute \(handle)")
+                log(level: .debug, from: self, "\(handle) result: \(result)")
+
                 guard let task = task else {
-                    log(from: self, "task.execute => task dead")
+                    log(from: self, "task.execute => task for \(handle) died")
```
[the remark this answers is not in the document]
aliak00: defaults to `info`

Source/Core/Logger/Logger.swift
aliak00: I would hope the destructor of DispatchGroup released it's resources. leave afaik just decrements a counter but I'm not sure of the implementation details. Couldn't find docs warning against anything weird here either. You know something I don't? 😛

Source/Core/Logger/Logger.swift
aliak00: ahhh my head started hurting when i tried !!

Source/Manager/User/User.swift
aliak00: levels default to info 👍

Source/Core/Logger/Logger.swift
Torsph: Only think I know is this:
```
 Calling this function indicates another block has joined the group through
 a means other than dispatch_group_async(). Calls to this function must be
 balanced with dispatch_group_leave().
```

Don't know what they mean with _must_ in:
 _Calls to this function must be balanced with dispatch_group_leave()_

Torsph: From this link https://github.com/apple/swift-corelibs-libdispatch/blob/master/dispatch/group.h#L237

PR #138, opened 2018-06-07
Fixed corner radius theming

### LINKED ISSUE #126: Theming the corner radius of buttons also effects input fields
They shouldn't.

aliak00:
Fixes #126

### DIFF
--- a/Source/UI/IdentityUITheme.swift
+++ b/Source/UI/IdentityUITheme.swift
@@ -15,12 +15,19 @@ public struct IdentityUITheme {
     /// The logo that's displayed on the identifier screen
     public var titleLogo: UIImage?
 
-    ///
+    /// Controls spacings and curvatures of views
     public struct Geometry {
-        /// Default curved corner radius of controls
-        public var cornerRadius: CGFloat = 4
-        /// Default spacing between views that form a grouping
+        /// Spacing between views that form a grouping
         public var groupedViewSpacing: CGFloat = 8
+        /// Spacing for in between titles and their associated input fields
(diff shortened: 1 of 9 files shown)


Source/UI/IdentityUITheme.swift
```diff
@@ -15,12 +15,19 @@ public struct IdentityUITheme {
...
+    /// Controls spacings and curvatures of views
     public struct Geometry {
-        /// Default curved corner radius of controls
-        public var cornerRadius: CGFloat = 4
-        /// Default spacing between views that form a grouping
+        /// Spacing between views that form a grouping
         public var groupedViewSpacing: CGFloat = 8
+        /// Spacing for in between titles and their associated input fields
+        public var titleViewSpacing: CGFloat = 4
+
```
Torsph: Maybe unneeded space?

Source/UI/Screens/Error/ErrorViewController.swift
```diff
@@ -93,6 +93,7 @@ class ErrorViewController: IdentityUIViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
         self.sheetBackgroundView.layer.cornerRadius = self.theme.geometry.cornerRadius
```
Torsph: Could maybe be better to use a `UIBezierPath()` if you want smoother corners..
https://twitter.com/nathangitter/status/992842150500069376

aliak00: Nice, I switch to bezier everywhere but in some places it was much worse:

image VS image

Becomes a bit more complicated to do if you want to add borders apparently.

PR #139, opened 2018-06-07
Fixed: custom image checkbox scaling

### LINKED ISSUE #123: UI: Setting a custom checkbox image does not work well
Trying to set a custom checkbox image results in image-resize and the remember me text being cut off.

checkboxchecked
Testing with this image.

screen shot 2018-05-25 at 12 52 32
Results in this.

aliak00:
Fixes #123

### DIFF
--- a/Source/UI/Views/Checkbox.swift
+++ b/Source/UI/Views/Checkbox.swift
@@ -13,11 +13,14 @@ class Checkbox: UIButton, Themeable {
 
     func applyTheme(theme: IdentityUITheme) {
         self.theme = theme
+        self.heightAnchor.constraint(equalToConstant: 26).isActive = true
+        self.widthAnchor.constraint(equalToConstant: 26).isActive = true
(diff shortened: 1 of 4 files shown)


Source/UI/Views/Checkbox.swift
```diff
@@ -13,11 +13,13 @@ class Checkbox: UIButton, Themeable {
 
     func applyTheme(theme: IdentityUITheme) {
         self.theme = theme
+        self.heightAnchor.constraint(equalToConstant: 26).isActive = true
+        self.widthAnchor.constraint(equalToConstant: 26).isActive = true
         self.setImage(theme.icons.checkedBox, for: .selected)
         self.setImage(theme.icons.uncheckedBox, for: .normal)
         self.setTitle(nil, for: UIControlState.normal)
         self.tintColor = self.stateColor
-        self.translatesAutoresizingMaskIntoConstraints = true
```
Torsph: should this still be here? But be set to false?

PR #140, opened 2018-06-07
Scopes

aliak00:


Example/StatusViewController.swift
```diff
@@ -61,7 +61,7 @@ class StatusViewController: UIViewController {
     }
 
     @IBAction func didClickPasswordlessPhoneLogin(_: Any) {
-        UIApplication.identityUI.presentIdentityProcess(from: self, loginMethod: .phone)
+        UIApplication.identityUI.presentIdentityProcess(from: self, loginMethod: .phone, scopes: ClientConfiguration.current.scopes)
```
Torsph: Should the scopes be injected into the controller instead of using a singleton?

Source/Manager/IdentityManager.swift
```diff
@@ -259,7 +262,7 @@ public class IdentityManager: IdentityManagerProtocol {
             connection: identifier.connection,
             code: oneTimeCode,
             passwordlessToken: passwordlessToken,
-            scope: ["openid"]
+            scope: scopes + IdentityManager.defaultScopes
```
Torsph: Could you end up with getting duplicates here? Does it matter if you get duplicates?

Source/Manager/IdentityManager.swift
```diff
@@ -364,7 +368,7 @@ public class IdentityManager: IdentityManagerProtocol {
             refreshToken: nil,
             username: username.normalizedString,
             password: password,
-            scope: scopes + ["openid"]
+            scope: scopes + IdentityManager.defaultScopes
```
Torsph: Same here.. Duplicates..

Example/StatusViewController.swift
aliak00: Ah, that's just an extension for the example app.

Source/Manager/IdentityManager.swift
aliak00: Hmm, good point. I fix.

PR #141, opened 2018-06-07
Added email to user profile

### LINKED ISSUE #15: Get primary email address from "email" not from "emails"
It should be the same, and is safer incase "primary" tag is missing from the array of emails.

See `UserProfile` model under the IdentityAPI for where to start from.

aliak00:
The primary email in the list of emails is not always there. This
one is more guaranteed to be there always I think

Fixes #15

### DIFF
--- a/Source/Manager/IdentityAPI/Models/UserProfile.swift
+++ b/Source/Manager/IdentityAPI/Models/UserProfile.swift
@@ -19,8 +19,10 @@ public struct UserProfile: JSONParsable {
     public var displayName: String?
     ///
     public var birthday: Birthdate?
-    ///
+    /// This is the email address that is marked as primary (there may or may not be one)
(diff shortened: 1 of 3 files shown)


Source/Manager/IdentityAPI/Models/UserProfile.swift
```diff
@@ -40,6 +42,9 @@ public struct UserProfile: JSONParsable {
         if let string = try? data.string(for: "birthday") {
             self.birthday = Birthdate(string: string)
         }
+        if let email = try? data.string(for: "email") {
```
Torsph: Codable 🗡

aliak00: :D => https://github.schibsted.io/spt-identity/identity-sdk-ios/issues/503

PR #145, opened 2018-06-12
Added: isPersistant to User

### LINKED ISSUE #121: Feature request: Public access to isPersistent on User
Use case: 
We persist the e-mail of the user that logs in to use the `emailWithPrefilledValue` when showing the login UI. If the user does not tick the "Remember me"-checkbox we'd want to not do this. 
Currently there is no access to information about the user's choice after sign-in.

Proposal:
On the User object - expose the `isPersistent` in some way so SDK-users can check it after sign-in.

aliak00:
Fixes #121


Source/Manager/User/User.swift
```diff
@@ -70,7 +70,15 @@ public class User: UserProtocol {
 
     private let dispatchQueue = DispatchQueue(label: "com.schibsted.identity.User", attributes: [])
     private var _tokens: TokenData?
-    private var isPersistent = false
+    private var _isPersistent = false
+
+    /**
+     Tells you if the user is persistent or not. A persistent user is one where the tokens are
+     stored in the keycchain.
```
Torsph: typo

Source/Manager/User/User.swift
```diff
@@ -70,7 +70,15 @@ public class User: UserProtocol {
...
+     */
+    public var isPersistent: Bool {
```
Torsph: Why do you need two versions of the isPersistent variable?

aliak00: I don't :p will fix!

aliak00:
On hold till thread safety is added in.

Issue #147, opened 2018-06-13
Nil environment in ClientConfiguration results in `deployStage=dev` in PulseTrackingEventsHandler  


eriksundin:
At Blocket we handle getting an updated clientID/clientSecret/serverURL for SPiD from our backend in a configuration call that is made on startup of the app. We trust this updated information blindly.

As a result of the response of that call we init the `ClientConfiguration` with a `serverURL` rather than a set environment. This will result in the `environment` being `nil` in the configuration.
Later on in SDK Pulse tracking, the `environment` is used to determine the `deployStage`, and an empty environment will default to `dev`. This is what happens for us currently in production, all events end up in `dev`.

We will work around this on our side. But if the tracker will keep relying on the `environment` variable it would be good if SDK users could set it even if `ClientConfiguration` is created with a custom URL.

Examples:
* The `ClientConfiguration` could do a best-effort to set environment by url-matching the given URL in the init method.
* You can pass environment with the serverURL in the init method.

Passing an optional environment with the customURL might be the easiest one.

PR #148, opened 2018-06-13
Changed: ClientConfiguration tries best effort to determine environme…

aliak00:
…nt from serverURL

Just matches against expected values and set the enviornment if it is
a known one

Fixes: #147

### DIFF
--- a/Source/Manager/ClientConfiguration.swift
+++ b/Source/Manager/ClientConfiguration.swift
@@ -68,20 +68,35 @@ public struct ClientConfiguration {
         /// payment.schibsted.no
         case norway
 
-        fileprivate func loadConfiguration() -> (serverURL: URL, trackingProvider: String?) {
+        fileprivate static func loadConfigurationData() -> [String: [String: String]] {
             guard let configFile = Bundle(for: IdentityManager.self).path(forResource: "Configuration", ofType: "plist") else {
                 preconditionFailure("Configuration.plist file not found in bundle")
             }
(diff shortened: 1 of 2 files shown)


Source/Manager/ClientConfiguration.swift
```diff
@@ -68,21 +68,36 @@ public struct ClientConfiguration {
         /// payment.schibsted.no
         case norway
 
-        fileprivate func loadConfiguration() -> (serverURL: URL, trackingProvider: String?) {
+        fileprivate static func loadConfigurationData() -> [String: [String: String]] {
```
Torsph: Does this need to be static? Or could it just be a lazy var or something?
static vars is created on app startup, so the more static vars the slower the app launch.

Source/Manager/ClientConfiguration.swift
```diff
@@ -68,21 +68,36 @@ public struct ClientConfiguration {
...
+            let config = Environment.loadConfigurationData()
             let env = String(describing: self)
             guard let serverURL = URL(string: config[env]?["url"] ?? "about:blank") else {
                 preconditionFailure("Could not load url from configuration environment: \(env)")
             }
             let trackingProviderComponent = config[env]?["provider"]
             return (serverURL, trackingProviderComponent)
         }
+
+        fileprivate static func dataForServerURL(_ serverURL: URL) -> (environment: Environment, trackingProvider: String?)? {
```
Torsph: same here.

PR #149, opened 2018-06-13
Added: skip button to UI with delegate to control

### LINKED ISSUE #143: Add optional skip link on Identification screen
Background: AB (and other subscription apps) will need to allow purchase of subscription in-app w/o signing up so that Apple can approve their app. They need to do these changes before July 1.

Solution:
- [ ] Add an optional Skip link on the Identification screen (clients will decide if they want to show it or not)
- [ ] Once the Skip link is clicked, close the identification screen
- [ ] If the user click the link, we need to provide a way for the client to know that the user tapped it. This is so they can provide their own customized action after the screen is closed.

Mock: 
- It's in Android but it should have the same principle as how do we do the "Forgot Password" on Password screen for Login

image 18

aliak00:
Fixes #143

### DIFF
--- a/Source/UI/IdentityUI.swift
+++ b/Source/UI/IdentityUI.swift
@@ -292,6 +292,8 @@ public class IdentityUI {
         case .cancel:
             self.configuration.tracker?.loginID = nil
             uiResult = .canceled
+        case .skip:
+            uiResult = .skipped
         case .onlyDismiss:
             uiResult = nil
         case let .failure(error):
@@ -339,6 +341,7 @@ extension IdentityUI: FlowCoordinator {
         case cancel
         case onlyDismiss
         case failure(ClientError)
+        case skip
     }
 
     func start(input: Input, completion: @escaping (Output) -> Void) {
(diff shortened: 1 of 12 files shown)


Source/UI/IdentityUI.swift
```diff
@@ -471,6 +474,19 @@ extension IdentityUI {
             case .back:
                 // First screen, `back` cancels the flow.
                 completion(.cancel)
+            case .skip:
+                if let topViewController = self?.navigationController.topViewController, let delegate = self?.delegate {
```
Torsph: This looks more like a guard :)

Source/UI/IdentityUIDelegate.swift
```diff
@@ -48,10 +60,24 @@ public protocol IdentityUIDelegate: class {
      this method.
      */
     func willPresent(flow: LoginMethod.FlowVariant) -> LoginFlowDisposition
+    /**
+     This will be called whent he user presses the skip button on a skippable UI flow
```
Torsph: typo...

Source/UI/IdentityUI.swift
aliak00: aye!

PR #152, opened 2018-06-19
Added login will succeed disposition

aliak00:


Source/UI/IdentityUI.swift
```diff
@@ -274,21 +274,23 @@ public class IdentityUI {
...
         }
 
         let uiResult: IdentityUIResult?
+        var shouldPersistUser = false
         switch output {
-        case let .success(user):
+        case let .success(user, persistUser):
             self.configuration.tracker?.engagement(.network(.done))
             uiResult = .completed(user)
+            shouldPersistUser = persistUser
```
Torsph: Why can we not just store the tokens here? Could the user still cancel the login process because of requiredFields or something? Maybe its that you are fixing now?

Source/UI/IdentityUI.swift
```diff
@@ -300,27 +302,64 @@ public class IdentityUI {
...
+
+            // This is no more the currently presented login flow
+            IdentityUI.presentedIdentityUI = nil
+
+            if self?.navigationController.presentingViewController != nil {
+                // It might be that `IdentityUIViewController.endLoading()` has been called just before getting here, in case the result of a networking
+                // operation caused the flow to end. The `endLoading()` method will then trigger a `view.isUserInteractionEnabled = true`, which would cause the
+                // keyboard to show up again during the dismiss animation, resulting in a very weird and funky UI glitch. In order to avoid that, we force an
+                // `endEditing()` on the topmost view (if any) before starting the view dismissing.
+                self?.navigationController.topViewController?.view.endEditing(true)
```
Torsph: Is this an issue that comes from some other kind of bug or issue? And if it is, should that be fixed instead?

Source/UI/IdentityUI.swift
```diff
@@ -300,27 +302,64 @@ public class IdentityUI {
...
+            }
+        }
+
+        if let delegate = self.delegate, case let .completed(user)? = uiResult {
+            delegate.willSucceed(with: user, on: self.navigationController.topViewController) { disposition in
+                switch disposition {
+                case .continue:
+                    dismissFlow()
+                case let .failed(title: title, message: message):
+                    if self.navigationController.presentingViewController == nil {
```
Torsph: Why not present the navigationController and viewController before start doing anything? Or is there some part of the flow im missing.

Source/UI/IdentityUIDelegate.swift
```diff
@@ -70,6 +85,21 @@ public protocol IdentityUIDelegate: class {
...
+     UI what to do next.
+
+     The view controller passed in is whatever is being shown at the time. This can be nil in some cases where the
+     login flow did not need to fire up a flow (when validating a signing deep link when launching the app for e.g.)
+
+     - parameter result: the `User` that will be given to `didFinish`
+     - parameter topViewController: The currently shown top view controller
+     - parameter done: call this to say you're done so the UI can continue
+     */
+    func willSucceed(with user: User, on topViewController: UIViewController?, done: @escaping (LoginWillSucceedDisposition) -> Void)
```
Torsph: For some reason I expected this to be called `willFinish` since you talked about `didFinish` in the documentation.

Source/UI/IdentityUI.swift
aliak00: Ya, perhaps, this should be looked in to, @lrossi had the original code here so maybe he has some insight?

Source/UI/IdentityUI.swift
aliak00: The patch gives a chance to client to do one last thing before finalizing. So they can choose to reset the flow, and then ya, not persisting.

Source/UI/IdentityUI.swift
aliak00: We might not need to if we come from a deep link to verify the auth code and the client says we are all good, so continue the flow.

Source/UI/IdentityUIDelegate.swift
aliak00: Hmm, I see I see.

PR #157, opened 2018-06-25
Cache terms

aliak00:


Source/Manager/SDKConfiguration/SDKConfiguration+AgreementsCache.swift
```diff
@@ -0,0 +1,81 @@
...
+        /// Maximum number of days agreements status should be cached
+        public let maxDays: UInt32
+
+        func load(forUserID userID: String) -> Agreements? {
+            guard self.isOn, let data = Settings.value(forKey: Keys.agreementsData) as? JSONObject else {
+                return nil
+            }
+            do {
+                let storedID = try data.string(for: Keys.Data.userID)
+                guard storedID == userID else {
```
Torsph: Any reason to not combine the `guards` here? One test on each line?

Source/Manager/SDKConfiguration/SDKConfiguration+AgreementsCache.swift
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        func store(_ agreements: Agreements, forUserID userID: String) {
+            guard self.isOn else {
+                return
+            }
+            let now = Date()
+            let hoursPerDay: UInt32 = 24
+            let number = Int(arc4random_uniform(UInt32(self.maxDays - self.minDays) * hoursPerDay) + self.minDays * hoursPerDay)
+            guard let later = Calendar.current.date(byAdding: .hour, value: number, to: now) else {
```
Torsph: later dude 🦆

Torsph: Smart to use the Calendar functions todo this 👍 No need to think about day wrapper, daylight saving etc.

Source/Manager/SDKConfiguration/SDKConfiguration+AgreementsCache.swift
aliak00: Just readability. Also might need try? but then I can't log the throw if it throws

Torsph: The try-catch is ok.. Was thinking more about the 3 guards :)

PR #174, opened 2018-07-30
Phone number AND user loading adjustments

aliak00:
* Added phone number field to UserProfile
* And updates Example app which resulted in slight changes to User keychain loading/unloading

### DIFF
--- a/Example/ProfileViewController.swift
+++ b/Example/ProfileViewController.swift
@@ -19,7 +19,7 @@ class ProfileViewController: UIViewController {
         guard let familyName = familyNameTextField.text else { return }
         let profile = UserProfile(givenName: givenName, familyName: familyName)
 
-        UIApplication.identityManager.currentUser.profile.update(profile) { result in
+        UIApplication.currentUser.profile.update(profile) { result in
             switch result {
             case .success:
                 print("profile.update success!")
--- a/Example/StatusViewController.swift
+++ b/Example/StatusViewController.swift
@@ -6,11 +6,20 @@
 @testable import SchibstedAccount
 import UIKit
 
-extension StatusViewController: IdentityManagerDelegate {
-    // IdentityManagerDelegate
-    func userStateChanged(_: UserState) {
-        print("Login state changed to: \(self.isUserLoggedIn)")
-        self.updateFromCurrentUser()
+extension StatusViewController: UserDelegate {
+    func user(_ user: User, didChangeStateTo state: UserState) {
+        print("user \(user) changed state to \(state)")
+        switch state {
+        case .loggedIn:
+            // This can only happen with headless login because a visual login results in IdentityUIDelegate.didFinish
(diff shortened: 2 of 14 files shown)


Example/ProfileViewController.swift
```diff
@@ -19,7 +19,7 @@ class ProfileViewController: UIViewController {
         guard let familyName = familyNameTextField.text else { return }
         let profile = UserProfile(givenName: givenName, familyName: familyName)
 
-        UIApplication.identityManager.currentUser.profile.update(profile) { result in
+        UIApplication.currentUser.profile.update(profile) { result in
```
Torsph: Looks a bit weird that you are updating `profile` with a new `profile`. Maybe its just a naming issue, but it looks like you could do `UIApplication.currentUser.profile = profile`

Example/StatusViewController.swift
```diff
@@ -6,11 +6,20 @@
...
-    func userStateChanged(_: UserState) {
-        print("Login state changed to: \(self.isUserLoggedIn)")
-        self.updateFromCurrentUser()
+extension StatusViewController: UserDelegate {
+    func user(_ user: User, didChangeStateTo state: UserState) {
+        print("user \(user) changed state to \(state)")
+        switch state {
+        case .loggedIn:
+            // This can only happen with headless login because a visual login results in IdentityUIDelegate.didFinish
+            UIApplication.currentUser = UIApplication.identityManager.currentUser
```
Torsph: Why do you have two currentUsers? Is it because of a delegate issue?

Source/Manager/UserTokensStorage.swift
```diff
@@ -66,4 +66,19 @@ struct UserTokensStorage {
             log(level: .error, from: self, "error removing keychain for user: \(error)", force: true)
         }
     }
+
+    func clearAll() throws {
+        let keychain = UserTokensKeychain()
+        keychain.removeAllTokens()
+
+        do {
+            try keychain.saveInKeychain()
```
Torsph: To next time. What is the difference between UserTokensKeychain and SPiDKeychainWrapper?

Example/ProfileViewController.swift
aliak00: Yeah, I think could've been better named. If you make it a computed setter then you need async await functionality to get async results :p

Example/StatusViewController.swift
aliak00: Because the example app uses two login methods which official is not recommended. I.e. apps should only use the UI to login, but the example app is also a playground to display all the features so it uses the headless as well. So we have two `IdentityManager` objects around.

Source/Manager/UserTokensStorage.swift
aliak00: SPiDKeychainWrapper reimplements the keychain how the old SPID SDK used to do it (copy pasted from there) and UserTokensKeychain is the current keychain.

PR #175, opened 2018-07-31
Added: User.loadLast

aliak00:


Source/Manager/User/User+load.swift
```diff
@@ -0,0 +1,17 @@
...
+// Licensed under the terms of the MIT license. See LICENSE in the project root.
+//
+
+import Foundation
+
+extension User {
+    /**
+     Loads the last user that was persisted to the keychain
+     */
+    public static func loadLast(withConfiguration clientConfiguration: ClientConfiguration) -> User {
```
Torsph: Is the `loadLast` just to be more specific that you would just load the last user that was stored?

aliak00: Ya, and also because now to get the last loaded use is quite non-obvious:
```
let lastUser  = IdentityManager(clientConfiguration: .configuration).currentUser
```

PR #177, opened 2018-08-03
added phone number required field

aliak00:


Source/Manager/Identifiers/PhoneNumber.swift
```diff
@@ -85,35 +79,42 @@ public struct PhoneNumber: IdentifierProtocol {
...
+        if string.hasPrefix("+") {
+            string.remove(at: string.startIndex)
+        } else if string.hasPrefix("00") {
+            string.removeFirst(2)
+        } else {
+            return nil
         }
 
-        guard decimalDigitCharacters.count > 0 else {
+        guard string.count > 0 && string.rangeOfCharacter(from: CharacterSet.decimalDigits.inverted) == nil else {
```
Torsph: Do you need the `string.count > 0`? Looks like it gives the same result without it.

aliak00: Si! incase only + or 00 is passed in

PR #180, opened 2018-08-06
Fixed phone number and error

aliak00:


Source/UI/Coordinators/CompleteProfileCoordinator.swift
```diff
@@ -267,7 +267,11 @@ extension CompleteProfileCoordinator {
 extension CompleteProfileCoordinator {
     private func handle(error: ClientError) -> Bool {
         if self.presentedViewController is TermsViewController || self.presentedViewController is RequiredFieldsViewController {
-            self.present(error: error)
+            if let vc = self.presentedViewController as? RequiredFieldsViewController, case .alreadyRegistered = error {
+                self.presentError(description: vc.viewModel.string(for: .numberInvalid))
```
ksimons: Huh? Why when it's an alreadyRegistered error do we show a numberInvalid error?

PR #184, opened 2018-08-08
Tests

aliak00:


Tests/Core/JSONTests.swift
```diff
@@ -0,0 +1,71 @@
...
+//
+
+import Nimble
+import Quick
+@testable import SchibstedAccount
+
+class JSONTests: QuickSpec {
+
+    override func spec() {
+        it("Should convert to Data and conver back correctly") {
```
ksimons: Typo.

Tests/UserProfileSessionTests.swift
```diff
@@ -47,5 +47,24 @@ class UserProfileSessionTests: QuickSpec {
                 }
             }
         }
+
+        describe("Required fields") {
+            it("should be there") {
```
ksimons: You use a capital S everywhere else 😛

PR #185, opened 2018-08-09
Change: swift 4.1 min

aliak00:


Source/Manager/Identifiers/EmailAddress.swift
```diff
@@ -54,3 +54,9 @@ public struct EmailAddress: IdentifierProtocol {
         return identifier
     }
 }
+
+extension EmailAddress: Equatable {
+    public static func == (lhs: EmailAddress, rhs: EmailAddress) -> Bool {
+        return lhs.normalizedString == rhs.normalizedString
+    }
+}
```
fespinoza: do you need this?

it seems the only property of `EmailAddress` is `emailAddress` which is a string that is already `Equatable`, then you just declare `EmailAddress: Equatable` you don't need to define `==` then

Tests/Core/JWTTests.swift
```diff
@@ -53,7 +53,7 @@ class JWTTests: QuickSpec {
                 expect { try proxy.toJSON(string: "blah.bl{ah.blah") }.to(throwError(JWTHelperError.componentDecodingError("")))
             }
 
-            fit("Should fail on invalid json in second jwt component") {
```
fespinoza: what is fit?

Source/Manager/Identifiers/EmailAddress.swift
aliak00: Ah nice catch! Thanks!

Tests/Core/JWTTests.swift
aliak00: It makes "only this test" run :p somehow got in from a previous commit.

PR #186, opened 2018-08-10
Pass in the presentingViewController instead of topViewController

aliak00:
Since a top view controller is possibly not actually presented


aliak00:
Will not release this feature right now.

PR #188, opened 2018-08-17
Changed: Store tokens with accessmode kSecAttrAccessibleAfterFirstUnlock to allow background access.

eriksundin:
**Problem**

If a killed app receives a push notifications with `content-available` set to `1`, and the app allows background mode, the app will be launched in the background by the system. 
If the device is locked when this happens - reads to keychain values with  `kSecAttrAccessibleWhenUnlocked` will fail with an error `Access to item attempted while keychain is locked.`.

If the app is relying on the state of the User logged in state as part of the launch of the app - that state will then signal that there are no access tokens stored, meaning the user is signed out.

We've been experiencing this problem at Blocket with users getting signed out, as they receive push notifications for messages.

**Proposed solution**

Rather than storing values to only be read while the device is unlocked, use the  `kSecAttrAccessibleAfterFirstUnlock` instead. As described by Apple:

> After the first unlock, the data remains accessible until the next restart. This is recommended for items that need to be accessed by background applications. Items with this attribute migrate to a new device when using encrypted backups.

This pull request does this change. And also changes the way duplicate values are handled.
In the previous code, the `kSecAttrAccessGroup` was included in attributes used for read, write and delete. When changing the value however, this will fail when you try to delete the value as the `kSecAttrAccessGroup` will differ from the previously stored value.

Instead, here, the deletion of a "duplicate" is moved to `saveInKeychain` where you try to delete the value first before inserting the new one. The attributes for deletion will not include the `kSecAttrAccessGroup`.

### DIFF
--- a/Source/Manager/SwiftKeychain.swift
+++ b/Source/Manager/SwiftKeychain.swift
@@ -48,7 +48,7 @@ protocol KeychainItemType {
 
 extension KeychainItemType {
     var accessMode: String {
-        return String(kSecAttrAccessibleWhenUnlocked)
+        return String(kSecAttrAccessibleAfterFirstUnlock)
     }
 
     var accessGroup: String? {
@@ -62,6 +62,7 @@ extension KeychainItemType {
         let archivedData = NSKeyedArchiver.archivedData(withRootObject: dataToStore)
 
         itemAttributes[String(kSecValueData)] = archivedData
+        itemAttributes[String(kSecAttrAccessible)] = accessMode
 
         if let group = accessGroup {
             itemAttributes[String(kSecAttrAccessGroup)] = group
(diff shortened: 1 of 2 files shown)


aliak00:
Wow, quite a catch. I've just pinged "the security guy" to make sure there're no gotchas here, but this seems fine to me otherwise. Thanks!! 👍

aliak00:
```
[14:36:02]: ▸ SchibstedAccountTests.UserTokensKeychainTests
[14:36:02]: ▸ Saving__should_remove_from_old_format_and_store_in_new_format, expected to throw error <Did not find key >, got <Error Domain=UserTokensKeychainTests.loadWithPreviousKeychainLayout Code=-25300 "(null)">
[14:36:02]: ▸ /Users/travis/build/schibsted/account-sdk-ios/Tests/UserTokensKeychainTests.swift:144
[14:36:02]: ▸ ```
[14:36:02]: ▸ 
[14:36:02]: ▸                 expect { try loadWithPreviousKeychainLayout() }.to(throwError(JSONError.noKey("")))
[14:36:02]: ▸             }
```

Do you get this locally as well?

aliak00:
🍪

PR #192, opened 2018-09-04
Fixed: setting new user tokens

aliak00:
We were clearning even if there were no new tokens to set so the
old keychain tokens would go baboom


aliak00:
@Astach said it was perfect.

PR #216, opened 2018-09-21
Added debug option to force refreshing of tokens

aliak00:


Source/Manager/TaskManager/TaskManager.swift
```diff
@@ -77,6 +77,21 @@ class TaskManager {
                         DispatchQueue.main.async {
                             completion?(result)
                         }
+
+                        if SDKConfiguration.shared.invalidateteAuthTokenAfterSuccessfullRequest {
```
minarikg: I would protect this part with #if DEBUG, just to be sure

Source/Manager/SDKConfiguration/SDKConfiguration.swift
```diff
@@ -36,4 +36,16 @@ public class SDKConfiguration {
...
     }
+
+    /**
+     Set this to debug access token and refreshing requests. If set to true then every successful request
+     will also invalidate the access token so that a refresh will be forces
+
+     Only settable in debug mode
+     */
+    #if DEBUG
+        public var invalidateteAuthTokenAfterSuccessfullRequest = false
```
Torsph: Why cannot this be set to `true` when its set to `true` in `Initialize.init` under debug?

Source/Manager/TaskManager/TaskManager.swift
```diff
@@ -77,6 +77,23 @@ class TaskManager {
                         DispatchQueue.main.async {
                             completion?(result)
                         }
+
+                        #if DEBUG
+                        if SDKConfiguration.shared.invalidateteAuthTokenAfterSuccessfullRequest {
```
Torsph: Should this be inside its own function to separate it more from the main code?

Source/Manager/SDKConfiguration/SDKConfiguration.swift
aliak00: Good question. My thought was that then ze clients will get the behavior by default when they are developing and I thought opt-in is nicer for debug stuff.

But ... 🤷‍♂️

Source/Manager/TaskManager/TaskManager.swift
aliak00: 🗡 😛

PR #224, opened 2018-10-10
Changed: Move token refresh retry limit to SDKConfiguration.

zamzterz:
As suggested during hand-over workshop.


Source/Manager/Protocols/UserAuthAPI.swift
```diff
@@ -7,8 +7,6 @@ import Foundation
 
 ///
 public protocol UserAuthAPI: class {
-    ///
-    var refreshRetryCount: Int? { get set }
```
aliak00: Swift has an `@deprecated` annotation that I think should be used here. I would suggest we keep this here and just deprecate it for the next version. It can be removed when we do a major version bump then.

zamzterz:
Superseded by #235.

PR #225, opened 2018-11-15
Changed: Make the entire verification code area clickable.

### LINKED ISSUE #210: Make tappable area around entering verification code BIGGER
You currently have to press on the first entry digit area to get the keyboard up and it's not a very big place to tap.

The whole verification code "area" should be tappable.

zamzterz:
Fix #210


aliak00:
👿

PR #239, opened 2019-01-02
Add retry on network error

epatel:
I am working on the current Aftonbladet app (the old code base). While fixing parts of the purchase flow for the non-logged-in-state I notice we were receiving a lot of "lost connection"/-1005 errors. The Google answer seem to be that the app drops "Keep-alive" connections (when leaving the app) which when returned becomes a lost connection. This becomes noticeable when leaving Aftonbladet to read and confirm the account creation confirmation email. When tapping confirm and returning to the app will show a "something went wrong" message, but works on a second try. To mitigate this behaviour I added a re-try step in IdentityAPI on just this error. Feel free to ping me in Schibsted slack for questions.

### DIFF
--- a/Source/Manager/IdentityAPI/IdentityAPI.swift
+++ b/Source/Manager/IdentityAPI/IdentityAPI.swift
@@ -21,8 +21,7 @@ class IdentityAPI {
 
         let block: ((Result<T, ClientError>) -> Void) = { [weak self] (result) in
             if retry > 0,
-                case .failure(let clientError) = result,
-                case .networkingError(let error) = clientError,
+                case .failure(.networkingError(let error)) = result,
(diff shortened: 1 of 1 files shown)


zamzterz:
This looks good in general, but would it be possible to add some tests/steps for how to reproduce it? To verify that it is actually due to `Keep-Alive` connections being dropped.

And if so, the retries should really be restricted to only happen when the error is 
[`NSURLErrorNetworkConnectionLost`](https://developer.apple.com/documentation/foundation/1508628-url_loading_system_error_codes/nsurlerrornetworkconnectionlost), since there are other error that can occur which shouldn't be retried.

zamzterz:
Replaced by #242 (for Travis sake).

PR #238, opened 2018-12-14
ID-2729 Show notice when creating new account

zamzterz:

PR #240, opened 2019-01-08
ID-2729 Show notice when creating new account

sa-mao:
Supersede #238.


zamzterz:
#238 was used instead.

PR #249, opened 2019-02-26
ID-2824: add biometric login

sa-mao:
Implement finger ID as authentication method in the new Identity flows for iOS

### DIFF
--- a/Source/InfoPlist.strings
+++ b/Source/InfoPlist.strings
@@ -1,0 +1,7 @@
+/* 
+  InfoPlist.strings
+  SchibstedAccount
+
+  Created by Oussama Sarhraoui on 2019-02-26.
+  Copyright © 2019 Schibsted. All rights reserved.
+*/
--- a/Source/InfoPlist.strings
+++ b/Source/InfoPlist.strings
@@ -1,7 +1,0 @@
-/* 
-  InfoPlist.strings
-  SchibstedAccount
-
-  Created by Oussama Sarhraoui on 2019-02-26.
-  Copyright © 2019 Schibsted. All rights reserved.
-*/
--- a/Source/UI/Coordinators/PasswordCoordinator.swift
+++ b/Source/UI/Coordinators/PasswordCoordinator.swift
@@ -37,9 +37,21 @@ class PasswordCoordinator: AuthenticationCoordinator, RouteHandler {
         context.evaluatePolicy(.deviceOwnerAuthenticationWithBiometrics, localizedReason: localizedReasonString) { success, _ in
             DispatchQueue.main.async {
                 if success {
-                    self.submit(password: password, for: input.identifier, on: input.loginFlowVariant, persistUser: true, scopes: input.scopes, completion: completion)
+                    self.submit(
+                        password: password,
+                        for: input.identifier,
+                        on: input.loginFlowVariant,
+                        persistUser: true,
+                        scopes: input.scopes,
+                        completion: completion
+                    )
                 } else {
-                    self.showPasswordView(for: input.identifier, on: input.loginFlowVariant, scopes: input.scopes,  completion: completion)
+                    self.showPasswordView(
+                        for: input.identifier,
+                        on: input.loginFlowVariant,
+                        scopes: input.scopes,
+                        completion: completion
+                    )
                 }
             }
         }
@@ -139,7 +151,11 @@ extension PasswordCoordinator {
             switch result {
             case let .success(currentUser):
                 self?.configuration.tracker?.loginID = currentUser.legacyID
-                self?.updatekeyChain(for: identifier,loginFlowVariant: loginFlowVariant, password: password) {
+                self?.updatekeyChain(
+                    for: identifier,
+                    loginFlowVariant: loginFlowVariant,
+                    password: password
+                ) {
                     self?.spawnCompleteProfileCoordinator(for: .signin(user: currentUser), persistUser: persistUser, completion: completion)
                 }
             case let .failure(error):
@@ -247,7 +263,7 @@ extension PasswordCoordinator {
             return nil
         }
     }
-    private func updatekeyChain(for identifier: Identifier, loginFlowVariant: LoginMethod.FlowVariant,  password: String, completion: @escaping () -> Void ) {
+    private func updatekeyChain(for identifier: Identifier, loginFlowVariant: LoginMethod.FlowVariant, password: String, completion: @escaping () -> Void ) {
         if #available(iOS 11.3, *) {
             let accessControl = SecAccessControlCreateWithFlags(kCFAllocatorDefault, kSecAttrAccessibleWhenPasscodeSetThisDeviceOnly, .userPresence, nil)
             var dictionary = [String: Any]()
--- a/Source/UI/Coordinators/PasswordCoordinator.swift
+++ b/Source/UI/Coordinators/PasswordCoordinator.swift
@@ -6,7 +6,7 @@
 import UIKit
 import LocalAuthentication
 
-fileprivate struct Constants {
+private struct Constants {
     static let BiometricsSecretsLabel = "com.schibsted.account.biometrics.secrets"
 }
 
@@ -260,8 +260,6 @@ extension PasswordCoordinator {
 
         var queryResult: AnyObject?
         let status = SecItemCopyMatching(query as CFDictionary, &queryResult)
-
-
         if status == noErr, let qresult =  queryResult as? Data, let password = String(data: qresult as Data, encoding: .utf8) {
             return password
         } else {
--- a/Source/UI/Coordinators/PasswordCoordinator.swift
+++ b/Source/UI/Coordinators/PasswordCoordinator.swift
@@ -260,6 +260,7 @@ extension PasswordCoordinator {
         }
         return true
     }
+
     private func clearKeyChain(for identifier: Identifier) {
         if canUseBiometrics() {
             var query = [String: Any]()
@@ -272,6 +273,7 @@ extension PasswordCoordinator {
             return
         }
     }
+
     private func updatekeyChain(
         for identifier: Identifier,
         loginFlowVariant: LoginMethod.FlowVariant,
--- a/Source/UI/Coordinators/PasswordCoordinator.swift
+++ b/Source/UI/Coordinators/PasswordCoordinator.swift
@@ -26,7 +26,11 @@ class PasswordCoordinator: AuthenticationCoordinator, RouteHandler {
         )
         // Are we allowed/able to use biometric login?
         let context = LAContext()
-        guard configuration.useBiometrics, #available(iOS 11.3, *), context.canEvaluatePolicy(.deviceOwnerAuthenticationWithBiometrics, error: nil) else {
+        guard configuration.useBiometrics,
+            #available(iOS 11.3, *),
+            context.canEvaluatePolicy(.deviceOwnerAuthenticationWithBiometrics, error: nil),
+            context.biometryType == .touchID
+        else {
             self.showPasswordView(for: input.identifier, on: input.loginFlowVariant, scopes: input.scopes, completion: completion)
             return
         }
@@ -271,15 +275,12 @@ extension PasswordCoordinator {
         completion: @escaping () -> Void
     ) {
         let context = LAContext()
-        guard #available(iOS 11.3, *), context.canEvaluatePolicy(.deviceOwnerAuthenticationWithBiometrics, error: nil) else {
+        guard #available(iOS 11.3, *),
+            context.canEvaluatePolicy(.deviceOwnerAuthenticationWithBiometrics, error: nil),
+            context.biometryType == .touchID
+        else {
             completion()
             return
-        }
-        let biometeryType: String
-        if context.biometryType == .faceID {
-            biometeryType = "Face ID"
-        } else {
-            biometeryType = "Touch ID"
         }
         let accessControl = SecAccessControlCreateWithFlags(kCFAllocatorDefault, kSecAttrAccessibleWhenPasscodeSetThisDeviceOnly, .userPresence, nil)
         var dictionary = [String: Any]()
@@ -304,9 +305,7 @@ extension PasswordCoordinator {
             )
             let message = viewModel.biometricsOnboardingMessage
                 .replacingOccurrences(of: "$0", with: configuration.appName)
-                .replacingOccurrences(of: "$1", with: biometeryType)
             let title = viewModel.biometricsOnboardingTitle
-                .replacingOccurrences(of: "$0", with: biometeryType)
 
             let alert = UIAlertController(title: title, message: message, preferredStyle: .alert)
             alert.addAction(UIAlertAction(title: "Enable", style: .default) { _ in
--- a/Source/UI/Coordinators/PasswordCoordinator.swift
+++ b/Source/UI/Coordinators/PasswordCoordinator.swift
@@ -24,7 +24,6 @@ class PasswordCoordinator: AuthenticationCoordinator, RouteHandler {
             loginFlowVariant: input.loginFlowVariant,
             localizationBundle: self.configuration.localizationBundle
         )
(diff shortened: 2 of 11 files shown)


Source/UI/Coordinators/PasswordCoordinator.swift
```diff
@@ -194,4 +229,72 @@ extension PasswordCoordinator {
...
+            var queryResult: AnyObject?
+            let status = withUnsafeMutablePointer(to: &queryResult) {
+                SecItemCopyMatching(query as CFDictionary, UnsafeMutablePointer($0))
+            }
+    
+            if status == noErr {
+                let password = String(data: queryResult as! Data, encoding: .utf8)!
+                print("Password: \(password)")
+                return password
+            } else {
```
aliak00: This could crash the app if the cast to `Data` fails. This can be avoided with optional binding:
```
if status == noErr, let qresult = queryResult as? Data {
    return password
} else {
```

Source/UI/Coordinators/PasswordCoordinator.swift
```diff
@@ -194,4 +229,72 @@ extension PasswordCoordinator {
...
+        }
+    }
+    private func updatekeyChain(for identifier: Identifier, loginFlowVariant: LoginMethod.FlowVariant,  password: String, completion: @escaping () -> Void ) {
+        if #available(iOS 11.3, *) {
+            let accessControl = SecAccessControlCreateWithFlags(kCFAllocatorDefault, kSecAttrAccessibleWhenPasscodeSetThisDeviceOnly, .userPresence, nil)
+            var dictionary = [String : Any]()
+            dictionary[kSecClass as String] = kSecClassGenericPassword
+            dictionary[kSecAttrLabel as String] = "com.schibsted.account.biometrics.secrets" as CFString
+            dictionary[kSecAttrAccount as String] = identifier.normalizedString as CFString
+            dictionary[kSecValueData as String] = password.data(using: .utf8)! as CFData
```
aliak00: Force unwrapping 🗡 😛

Source/UI/Coordinators/PasswordCoordinator.swift
```diff
@@ -194,4 +229,72 @@ extension PasswordCoordinator {
...
+            dictionary[kSecAttrAccessControl as String] = accessControl
+            
+            let identityNotFirstLogin = UserDefaults.standard.bool(forKey: "identity.isNotFirst.login")
```
aliak00: So the standard `UserDefaults` is shared between the whole app. So you just have to be aware of potential key clashes. There's a `Settings` object just for this purpose in the SDK, which also makes sure that every setting you put in to whatever "storage" the sdk is using (UserDefaults by default), the key is prepended with something like "com.schibsted.identity" and when you retrieve keys this is taken in to account. See the Core/Settings.swift file for details. Also, the `Settings` object is configurable and injectable by the tests as a bonus.

So this would be
```
if let value = Settings.value(forKey: "isFirstLogin") as? Bool, value  {
  // we have a value and it is true
} else {
  // We don't have a value OR it is false
}
```

Source/UI/Coordinators/PasswordCoordinator.swift
```diff
@@ -194,4 +229,72 @@ extension PasswordCoordinator {
...
+            if identityNotFirstLogin == false {
+                UserDefaults.standard.set(true, forKey: "identity.isNotFirst.login")
```
aliak00: There's a `Settings.setValue(...)` as well.

Source/UI/IdentityUIConfiguration.swift
```diff
@@ -19,6 +19,17 @@ public struct IdentityUIConfiguration {
     public let tracker: TrackingEventsHandler?
     /// This determines whether you want to allow the user to dismiss the login flow
     public let isCancelable: Bool
+    /// This will determin if you want to use biometric login.
+    public var useBiometrics: Bool {
+        get {
+            // This will return false if the key is not present.
+            let value = UserDefaults.standard.bool(forKey: "Identity.useBiometrics")
```
aliak00: Same this about using the `Settings` object.

aliak00:
PS: the CI failures are because of the force unwrappings apparently 😛. There's a formatting script as well in the scripts folder that uses swiftformat to ensure consistency. The CI will also fail if that is not done I believe.

Source/InfoPlist.strings
```diff
@@ -0,0 +1,7 @@
+/* 
+  InfoPlist.strings
+  SchibstedAccount
+
+  Created by Oussama Sarhraoui on 2019-02-26.
+  Copyright © 2019 Schibsted. All rights reserved.
+*/
```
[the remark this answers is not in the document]
sa-mao: Yes it is.

Source/UI/Coordinators/PasswordCoordinator.swift
sa-mao: I didn't know that! Thank you!

Source/UI/Coordinators/PasswordCoordinator.swift
```diff
@@ -14,7 +15,40 @@ class PasswordCoordinator: AuthenticationCoordinator, RouteHandler {
...
+                        }
+                    }
+                } else {
+                    self.showPasswordView(for: input.identifier, on: input.loginFlowVariant, scopes: input.scopes, completion: completion)
+                }
+            }
+            
+        } else {
+            self.showPasswordView(for: input.identifier, on: input.loginFlowVariant, scopes: input.scopes, completion: completion)
+        }
```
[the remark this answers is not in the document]
sa-mao: 👍

Source/UI/Coordinators/PasswordCoordinator.swift
```diff
@@ -14,7 +15,40 @@ class PasswordCoordinator: AuthenticationCoordinator, RouteHandler {
...
+                self.showPasswordView(for: input.identifier, on: input.loginFlowVariant, scopes: input.scopes,  completion: completion)
+            } else {
+                let context = LAContext()
+                let localizedReasonString = viewModel.biometricsPrompt + input.identifier.normalizedString
+                var authError: NSError?
+                if context.canEvaluatePolicy(.deviceOwnerAuthenticationWithBiometrics, error: &authError) {
+                    context.evaluatePolicy(.deviceOwnerAuthenticationWithBiometrics, localizedReason: localizedReasonString) { success, evaluateError in
+                        DispatchQueue.main.async {
+                            if success{
+                                self.submit(password: password!, for: input.identifier, on: input.loginFlowVariant, persistUser: true, scopes: input.scopes, completion: completion)
```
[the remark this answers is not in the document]
sa-mao: For your suggestion about `DispatchQueue.main.async` I tried it actually and didn't work.

Source/UI/IdentityUIConfiguration.swift
sa-mao: I can't use settings for defaults here. So I think I will skip this.

Source/UI/Coordinators/PasswordCoordinator.swift
```diff
@@ -14,7 +15,40 @@ class PasswordCoordinator: AuthenticationCoordinator, RouteHandler {
...
+        let viewModel = PasswordViewModel(
+            identifier: input.identifier,
+            loginFlowVariant: input.loginFlowVariant,
+            localizationBundle: self.configuration.localizationBundle
+        )
+        // Are we allowed to use biometric login?
+        if configuration.useBiometrics, #available(iOS 11.3, *) {
+            let password = self.getPasswordFromKeychain(for: input.identifier)
+            if password == nil {
+                self.showPasswordView(for: input.identifier, on: input.loginFlowVariant, scopes: input.scopes,  completion: completion)
```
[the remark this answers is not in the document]
sa-mao: Fixed!

Source/UI/Coordinators/PasswordCoordinator.swift
```diff
@@ -14,7 +15,40 @@ class PasswordCoordinator: AuthenticationCoordinator, RouteHandler {
...
+                    context.evaluatePolicy(.deviceOwnerAuthenticationWithBiometrics, localizedReason: localizedReasonString) { success, evaluateError in
+                        DispatchQueue.main.async {
+                            if success{
+                                self.submit(password: password!, for: input.identifier, on: input.loginFlowVariant, persistUser: true, scopes: input.scopes, completion: completion)
+                            } else {
+                                self.showPasswordView(for: input.identifier, on: input.loginFlowVariant, scopes: input.scopes,  completion: completion)
+                            }
+                        }
+                    }
+                } else {
```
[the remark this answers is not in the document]
sa-mao: fixed!

Source/UI/Coordinators/PasswordCoordinator.swift
sa-mao: Fixed!

Source/UI/Coordinators/PasswordCoordinator.swift
sa-mao: Fixed!

Source/UI/Coordinators/PasswordCoordinator.swift
```diff
@@ -194,4 +239,66 @@ extension PasswordCoordinator {
...
+        guard #available(iOS 11.3, *) else {
+            // Fallback to passsword login
+            return nil
+        }
+        var query = [String: Any]()
+        query[kSecClass as String] = kSecClassGenericPassword
+        query[kSecReturnData as String] = kCFBooleanTrue
+        query[kSecAttrAccount as String] = identifier.normalizedString as CFString
+        query[kSecAttrLabel as String] = "com.schibsted.account.biometrics.secrets" as CFString
+        query[kSecUseOperationPrompt as String] = "Please put your fingers on that button" as CFString
```
aliak00: Does this need localization?

Source/UI/Coordinators/PasswordCoordinator.swift
```diff
@@ -194,4 +239,66 @@ extension PasswordCoordinator {
...
+
+        var queryResult: AnyObject?
+        let status = withUnsafeMutablePointer(to: &queryResult) {
+            SecItemCopyMatching(query as CFDictionary, UnsafeMutablePointer($0))
+        }
+
+        if status == noErr, let qresult =  queryResult as? Data {
+            let password = String(data: qresult as Data, encoding: .utf8)!
```
aliak00: If the data is corrupted for any reason this would crash. I'd suggest you hoist the `let password = blah` part in to the `if` (variables unwrapped in an if are usable in the same if)

Source/UI/Coordinators/PasswordCoordinator.swift
```diff
@@ -194,4 +239,66 @@ extension PasswordCoordinator {
...
+        } else {
+            return nil
+        }
+    }
+    private func updatekeyChain(for identifier: Identifier, loginFlowVariant: LoginMethod.FlowVariant, password: String, completion: @escaping () -> Void ) {
+        if #available(iOS 11.3, *) {
+            let accessControl = SecAccessControlCreateWithFlags(kCFAllocatorDefault, kSecAttrAccessibleWhenPasscodeSetThisDeviceOnly, .userPresence, nil)
+            var dictionary = [String: Any]()
+            dictionary[kSecClass as String] = kSecClassGenericPassword
+            dictionary[kSecAttrLabel as String] = "com.schibsted.account.biometrics.secrets" as CFString
```
aliak00: I see this string is used in multiple places. So maybe make this a constant up in the file somewhere? I guess the convention is either:
```
private let kSecAttrLabelValue = "blah"
```
or
```
fileprivate struct Constants {
  static let kSecAttrLabelValue = "blah"
}
```

Source/UI/IdentityUIConfiguration.swift
aliak00: So I see it's because of the use of it in the default argument. Suggestion to not use `UserDefaults` as the default value. You can do this:
```
@@ -74,7 +73,7 @@ public struct IdentityUIConfiguration {
         theme: IdentityUITheme = .default,
         isCancelable: Bool = true,
         isSkippable: Bool = false,
-        useBiometrics: Bool = UserDefaults.standard.bool(forKey: "Identity.useBiometrics"),
+        useBiometrics: Bool? = nil,
         presentationHook: ((UIViewController) -> Void)? = nil,
         tracker: TrackingEventsHandler? = nil,
         localizationBundle: Bundle? = nil,
@@ -90,7 +89,7 @@ public struct IdentityUIConfiguration {
         if let appName = appName {
             self.appName = appName
         }
-        self.useBiometrics = useBiometrics
+        self.useBiometrics = useBiometrics ?? self.useBiometrics
     }
```

Now you can use `Settings` inside the `useBiometrics` property.

Source/UI/IdentityUIConfiguration.swift
```diff
@@ -89,11 +103,13 @@ public struct IdentityUIConfiguration {
      - parameter tracker: Required implementation of the trackinge events handler
      - parameter localizationBundle: If you have any custom localizations you want to use
      - parameter appName: If you want to customize the app name display in the UI
+     - parameter useBiometrics: If you want to enable biometrics authentication.
     */
     public func replacing(
         theme: IdentityUITheme? = nil,
         isCancelable: Bool? = nil,
         isSkippable: Bool? = nil,
+        useBiometrics: Bool? = UserDefaults.standard.bool(forKey: "Identity.useBiometrics"),
```
aliak00: This can be `nil` like the others and when you set the property it will keep updating of the storage in the setter.

Source/UI/Coordinators/PasswordCoordinator.swift
sa-mao: it seems like I never get a prompt here.

sa-mao: This needs further tests on a real device.

Source/UI/Coordinators/PasswordCoordinator.swift
```diff
@@ -194,4 +247,76 @@ extension PasswordCoordinator {
...
+        }
+    }
+    private func updatekeyChain(
+        for identifier: Identifier,
+        loginFlowVariant: LoginMethod.FlowVariant,
+        password: String,
+        completion: @escaping () -> Void
+    ) {
+        let context = LAContext()
+        guard #available(iOS 11.3, *),
```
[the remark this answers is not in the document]
sa-mao: No it also depends on if it's a first login.

Source/UI/Coordinators/PasswordCoordinator.swift
sa-mao: After a test on a real device it seems like I get a prompt there.

Source/UI/Coordinators/PasswordCoordinator.swift
```diff
@@ -112,8 +140,15 @@ extension PasswordCoordinator {
...
-                self?.spawnCompleteProfileCoordinator(for: .signin(user: currentUser), persistUser: persistUser, completion: completion)
+                self?.updatekeyChain(
+                    for: identifier,
+                    loginFlowVariant: loginFlowVariant,
+                    password: password
+                ) {
+                    self?.spawnCompleteProfileCoordinator(for: .signin(user: currentUser), persistUser: persistUser, completion: completion)
+                }
             case let .failure(error):
+                self?.clearKeyChain(for: identifier)
```
zamzterz: Nice catch!! 💯

Source/UI/IdentityUIConfiguration.swift
```diff
@@ -104,19 +125,30 @@ public struct IdentityUIConfiguration {
...
             appName: appName ?? self.appName
         )
     }
+
+    /**
+     Call this to enroll biometrics login
+
+    - parameter useBiometrics: If you want to enable biometrics authentication.
+    */
+    public func useBiometrics(_ useBiometrics: Bool) {
```
zamzterz: This function is redundant since `useBiometrics` is a public computed property implementing the same behavior. So all usages should be replaced with `configuration.useBiometrics = value`.

PR #253, opened 2019-03-27
Implement mobile device fingerprint endpoint

sa-mao:
Create a device fingerprint after each successful login.

### DIFF
--- a/Source/Manager/APITasks/UpdateUserDeviceTask.swift
+++ b/Source/Manager/APITasks/UpdateUserDeviceTask.swift
@@ -1,5 +1,5 @@
 //
-// Copyright 2011 - 2018 Schibsted Products & Technology AS.
+// Copyright 2011 - 2019 Schibsted Products & Technology AS.
 // Licensed under the terms of the MIT license. See LICENSE in the project root.
 //
 
@@ -13,13 +13,12 @@ class UpdateUserDeviceTask: TaskProtocol {
     }
 
     func execute(completion: @escaping NoValueCallback) {
-        guard let user = self.user, let tokens = user.tokens, let userID = tokens.anyUserID else {
+        guard let user = self.user, let tokens = user.tokens else {
             completion(.failure(.invalidUser))
             return
         }
 
         user.api.updateUserDevice(
-            userID: userID,
             oauthToken: tokens.accessToken,
             device: self.device
         ) { [weak self] result in
--- a/Source/Manager/APITasks/UpdateUserDeviceTask.swift
+++ b/Source/Manager/APITasks/UpdateUserDeviceTask.swift
@@ -1,0 +1,38 @@
+//
+// Copyright 2011 - 2018 Schibsted Products & Technology AS.
+// Licensed under the terms of the MIT license. See LICENSE in the project root.
+//
+
+class UpdateUserDeviceTask: TaskProtocol {
+    private weak var user: User?
+    private var device: UserDevice
+
+    init(user: User?, device: UserDevice) {
+        self.user = user
+        self.device = device
+    }
+
+    func execute(completion: @escaping NoValueCallback) {
+        guard let user = self.user, let tokens = user.tokens, let userID = tokens.anyUserID else {
+            completion(.failure(.invalidUser))
+            return
+        }
+
+        user.api.UpdateUserDevice(
+            userID: userID,
+            oauthToken: tokens.accessToken,
+            device: self.device
+        ) { [weak self] result in
+            guard self?.user != nil else {
+                completion(.failure(.invalidUser))
+                return
+            }
+            switch result {
+            case .success:
+                completion(.success(()))
+            case let .failure(error):
+                completion(.failure(ClientError(error)))
+            }
+        }
+    }
+}
--- a/Source/Manager/APITasks/UpdateUserDeviceTask.swift
+++ b/Source/Manager/APITasks/UpdateUserDeviceTask.swift
@@ -18,7 +18,7 @@ class UpdateUserDeviceTask: TaskProtocol {
             return
         }
 
-        user.api.UpdateUserDevice(
+        user.api.updateUserDevice(
             userID: userID,
             oauthToken: tokens.accessToken,
             device: self.device
--- a/Source/Manager/IdentityAPI/IdentityAPI.swift
+++ b/Source/Manager/IdentityAPI/IdentityAPI.swift
@@ -229,7 +229,7 @@ class IdentityAPI {
         requestWithRetries(router: .updateProfile(userID: userID), formData: formData, headers: [.authorization: oauthToken.bearer], completion: completion)
     }
 
-    func UpdateUserDevice(
+    func updateUserDevice(
         userID: String,
         oauthToken: String,
         device: UserDevice,
--- a/Source/Manager/IdentityAPI/IdentityAPI.swift
+++ b/Source/Manager/IdentityAPI/IdentityAPI.swift
@@ -229,6 +229,16 @@ class IdentityAPI {
         requestWithRetries(router: .updateProfile(userID: userID), formData: formData, headers: [.authorization: oauthToken.bearer], completion: completion)
     }
 
+    func UpdateUserDevice(
+        userID: String,
+        oauthToken: String,
+        device: UserDevice,
+        completion: @escaping ((Result<UserDevice, ClientError>) -> Void)
+    ) {
+        let formData = device.formData()
+        requestWithRetries(router: .devices(userID: userID), formData: formData, headers: [.authorization: oauthToken.bearer], completion: completion)
+    }
+
     func validateCode(clientID: String,
                       clientSecret: String,
                       identifier: String,
@@ -359,6 +369,8 @@ class IdentityAPI {
             return .tooManyRequests
         case let .object("ApiException", .string(string), 400) where string.contains("valid phone number"):
             return .invalidPhoneNumber
+        case let .object("ApiException", .string(string), 400) where string.contains("Invalid payload data"):
+            return .invalidDevicePayloadData
         case let .object("ApiException", .string(string), 302):
             return .alreadyRegistered(message: string)
         case let .object("ApiException", .string(string), 404) where string == "No access to that product.":
--- a/Source/Manager/IdentityAPI/IdentityAPI.swift
+++ b/Source/Manager/IdentityAPI/IdentityAPI.swift
@@ -232,7 +232,7 @@ class IdentityAPI {
     internal func updateUserDevice(
         oauthToken: String,
         device: UserDevice,
-        completion: @escaping ((Result<UserDevice, ClientError>) -> Void)
+        completion: @escaping ((Result<UserDeviceHash, ClientError>) -> Void)
     ) {
         let formData = device.formData()
         requestWithRetries(router: .devices, formData: formData, headers: [.authorization: oauthToken.bearer], completion: completion)
--- a/Source/Manager/IdentityAPI/IdentityAPI.swift
+++ b/Source/Manager/IdentityAPI/IdentityAPI.swift
@@ -230,13 +230,12 @@ class IdentityAPI {
     }
 
     internal func updateUserDevice(
-        userID: String,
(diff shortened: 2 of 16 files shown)


Source/Manager/IdentityAPI/IdentityAPI.swift
```diff
@@ -229,6 +229,16 @@ class IdentityAPI {
         requestWithRetries(router: .updateProfile(userID: userID), formData: formData, headers: [.authorization: oauthToken.bearer], completion: completion)
     }
 
+    func UpdateUserDevice(
```
aliak00: camel case

Source/Manager/IdentityAPI/Models/UserDevice.swift
```diff
@@ -0,0 +1,59 @@
...
+// Copyright 2011 - 2018 Schibsted Products & Technology AS.
+// Licensed under the terms of the MIT license. See LICENSE in the project root.
+//
+
+import Foundation
+
+/**
+ The user device data.
+
+ SeeAlso: ?
```
aliak00: seealso ?

Source/UI/Coordinators/PasswordCoordinator.swift
```diff
@@ -139,12 +139,24 @@ extension PasswordCoordinator {
...
+                    case .success:
+                        self?.updatekeyChain(
+                            for: identifier,
+                            loginFlowVariant: loginFlowVariant,
+                            password: password
+                        ) {
+                            self?.spawnCompleteProfileCoordinator(for: .signin(user: currentUser), persistUser: persistUser, completion: completion)
+                        }
+                    case let .failure(error):
+                       self?.present(error: error)
```
aliak00: Does the login process have to stop if the device can't be updated? Should the update device call be fire and forget?

Source/UI/IdentityUIConfiguration.swift
```diff
@@ -60,6 +60,13 @@ public struct IdentityUIConfiguration {
         }
     }
 
+    public var appVersion: String {
+        guard let name = Bundle.main.infoDictionary?[kCFBundleVersionKey as String] as? String else {
+            preconditionFailure("Could not fetch bundle name. Please set IdentityUIConfiguration.appName")
```
aliak00: Isn't this supposed to be something that much be changed in the app bundle and not the IdentityUIConfig object?

Source/Manager/IdentityAPI/Models/UserDevice.swift
```diff
@@ -0,0 +1,59 @@
...
+ */
+public struct UserDevice: JSONParsable {
+    ///
+    public var hash: String?
+    private var deviceId: String?
+    public var applicationName: String?
+    public var applicationVersion: String?
+    public var platform: String?
```
aliak00: Do all of these need to be public?

Source/UI/Coordinators/PasswordCoordinator.swift
sa-mao: No this is not supposed to fail.

Source/UI/IdentityUIConfiguration.swift
sa-mao: this is not supposed to be changed at all.

aliak00: The error message is off then isn't it?

Source/Manager/APITasks/UpdateUserDeviceTask.swift
```diff
@@ -0,0 +1,38 @@
+//
+// Copyright 2011 - 2018 Schibsted Products & Technology AS.
```
zamzterz: Copyright year should cover 2019 too I guess 🤷‍♂️

Source/Manager/IdentityAPI/Models/UserDevice.swift
```diff
@@ -0,0 +1,59 @@
+//
+// Copyright 2011 - 2018 Schibsted Products & Technology AS.
```
zamzterz: 2019? 🙂

Source/Manager/IdentityAPI/Router.swift
```diff
@@ -23,6 +23,7 @@ enum Router {
     case requiredFields(userID: String)
     case client(clientID: String)
     case product(userID: String, productID: String)
+    case devices(userID: String)
```
zamzterz: This no longer takes the `userID` as a route parameter (instead basing the user id on the included token).

Source/Manager/IdentityAPI/IdentityAPI.swift
```diff
@@ -229,6 +229,16 @@ class IdentityAPI {
         requestWithRetries(router: .updateProfile(userID: userID), formData: formData, headers: [.authorization: oauthToken.bearer], completion: completion)
     }
 
+    internal func updateUserDevice(
+        userID: String,
```
zamzterz: Since the `/devices` endpoint no longer has the user id as a route param this can be removed.

Source/Manager/Protocols/UserDeviceAPI.swift
```diff
@@ -0,0 +1,13 @@
+//
+// Copyright 2011 - 2018 Schibsted Products & Technology AS.
```
zamzterz: 2019

Source/Manager/User/User+Device.swift
```diff
@@ -0,0 +1,32 @@
+//
+// Copyright 2011 - 2018 Schibsted Products & Technology AS.
```
zamzterz: 2019

Source/Manager/IdentityAPI/Models/UserDevice.swift
```diff
@@ -0,0 +1,59 @@
...
+import Foundation
+
+/**
+ The user device data.
+
+ SeeAlso: https://techdocs.spid.no/types/device-fingerprint/
+ */
+public struct UserDevice: JSONParsable {
+    ///
+    private var hash: String?
```
zamzterz: I think this should be separated into two classes for stricter typing and separation of the request and respose for `/devices`:
1. User device data sent to `/devices`, (maybe `UserDeviceFingerprint`): `deviceId` (required), `platform` (required), `applicationName ` (required), `applicationVersion` (optional)
1. Device fingerprint hash returned in the response from `/devices` (maybe `UserDeviceHash`): `hash` (required)

Source/UI/Coordinators/PasswordCoordinator.swift
```diff
@@ -139,12 +139,19 @@ extension PasswordCoordinator {
...
             case let .success(currentUser):
                 self?.configuration.tracker?.loginID = currentUser.legacyID
-                self?.updatekeyChain(
-                    for: identifier,
-                    loginFlowVariant: loginFlowVariant,
-                    password: password
-                ) {
-                    self?.spawnCompleteProfileCoordinator(for: .signin(user: currentUser), persistUser: persistUser, completion: completion)
+                let device = UserDevice(
+                    hash: nil,
```
zamzterz: This `hash: nil` could be avoided by splitting `UserDevice` as suggested above.

Source/UI/Coordinators/PasswordCoordinator.swift
```diff
@@ -139,12 +139,19 @@ extension PasswordCoordinator {
...
+                    applicationName: self?.configuration.appName,
+                    applicationVersion: self?.configuration.appVersion
+                )
+                currentUser.device.update(device) { _ in
+                    self?.updatekeyChain(
```
zamzterz: Does this need to be nested as a callback? If possible I think it would make sense to move the `self?.updatekeyChain` out of the callback to avoid waiting on the device fingerprint request to be completed (since we don't really care about the result).

Source/UI/Coordinators/PasswordlessCoordinator.swift
```diff
@@ -85,7 +85,15 @@ extension PasswordlessCoordinator {
...
             case let .success(currentUser):
                 self?.configuration.tracker?.loginID = currentUser.legacyID
-                self?.spawnCompleteProfileCoordinator(for: currentUser, on: loginFlowVariant, persistUser: persistUser, completion: completion)
+                let device = UserDevice(
+                    hash: nil,
+                    applicationName: self?.configuration.appName,
+                    applicationVersion: self?.configuration.appVersion
+                )
+                currentUser.device.update(device) { _ in
+                    self?.spawnCompleteProfileCoordinator(for: currentUser, on: loginFlowVariant, persistUser: persistUser, completion: completion)
```
zamzterz: Same in `PasswordCoordinator`: Can this be moved outside the callback?

Source/Manager/IdentityAPI/IdentityAPI.swift
```diff
@@ -229,6 +229,16 @@ class IdentityAPI {
         requestWithRetries(router: .updateProfile(userID: userID), formData: formData, headers: [.authorization: oauthToken.bearer], completion: completion)
     }
 
+    internal func updateUserDevice(
```
zamzterz: Minor naming suggestion: Maybe `updateDeviceFingerprint` is more clear?

Source/Manager/IdentityAPI/IdentityAPI.swift
```diff
@@ -229,6 +229,16 @@ class IdentityAPI {
...
+        userID: String,
+        oauthToken: String,
+        device: UserDevice,
+        completion: @escaping ((Result<UserDevice, ClientError>) -> Void)
```
zamzterz: As mentioned below `UserDevice` should be split in two, and this would then become something like `Result<UserDeviceHash, ClientError>`.

Source/Manager/Protocols/UserDeviceAPI.swift
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the terms of the MIT license. See LICENSE in the project root.
+//
+
+import Foundation
+
+///
+public protocol UserDeviceAPI {
```
zamzterz: Why does this need to be `public`?

Source/Manager/User/User.swift
```diff
@@ -51,6 +51,9 @@ public class User: UserProtocol {
     /// Provides access to this user's profile data
     public internal(set) var profile: UserProfileAPI
 
+    /// Provides access to this user's device data
+    public internal(set) var device: UserDeviceAPI
```
zamzterz: Should this be `public`?

Source/Manager/IdentityAPI/Models/UserDevice.swift
```diff
@@ -0,0 +1,59 @@
...
+    private var applicationVersion: String?
+    private var platform: String?
+    ///
+    public init(
+        hash: String? = nil,
+        applicationName: String? = nil,
+        applicationVersion: String? = nil
+    ) {
+        self.hash = hash
+        self.deviceId = UIDevice.current.identifierForVendor!.uuidString
```
zamzterz: Is this the best identifier available on iOS? [Documentation](https://developer.apple.com/documentation/uikit/uidevice/1620059-identifierforvendor) says: "A different value is returned for apps on the same device that come from different vendors [...]", meaning it's only per app and not per device. Also is there some value we can use if `identifierForVendor` is `nil`, or does that not even happen? 🤔

Source/Manager/IdentityAPI/Models/UserDevice.swift
sa-mao: Here I tried to be consistent with the existing approach in `UserProfile`:
```    
func updateUserProfile(
        userID: String,
        oauthToken: String,
        profile: UserProfile,
        completion: @escaping ((Result<UserProfile, ClientError>) -> Void)
)
```
Notice that the response/request share the same model.

zamzterz: That endpoint (`/user/<userId>`) returns similar formatted data in the response as is sent in the request, basically echoing the stored user data for the given userId. So it makes more sense for that one, than the device fingerprint endpoint which returns completely different data in the response compared to the request.

PR #255, opened 2019-04-08
Update year in copyright notice.

zamzterz:


scripts/format.sh
```diff
@@ -4,7 +4,7 @@ set -e
 
 copyright="\
 \n\
-Copyright 2011 - 2018 Schibsted Products & Technology AS.\n\
+Copyright 2011 - 2019 Schibsted Products & Technology AS.\n\
```
sa-mao: Wouldn't be better to script the Year:
```
copyright="\
\n\
Copyright 2011 - `date +'%Y'` Schibsted Products & Technology AS.\n\
Licensed under the terms of the MIT license. See LICENSE in the project root.\n\
"
```

PR #256, opened 2019-04-24
Add possibility to inject bundle to ClientConfiguration

marcincs330:
I ran across the issue with the recent SchibstedAccount (2.1.0) version. `ClientConfiguration` init now uses `Bundle.main` directly and that makes it impossible to create an instance of it our test code. Precondition fails because tests don’t use main bundle. 

This my proposed solution, feel free to suggest anything smarter!

### DIFF
--- a/Source/Manager/ClientConfiguration.swift
+++ b/Source/Manager/ClientConfiguration.swift
@@ -136,7 +136,7 @@ public struct ClientConfiguration {
      - parameter appURLScheme: set your `appURLSceheme` here. Defaults to "spid-\(clientID)" if nil
      - parameter locale: Locale you want to use for requests - defaults to system
      */
-    public init(serverURL: URL, clientID: String, clientSecret: String, appURLScheme: String?, locale: Locale? = nil) {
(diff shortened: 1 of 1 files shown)


Source/Manager/ClientConfiguration.swift
```diff
@@ -136,7 +136,7 @@ public struct ClientConfiguration {
      - parameter appURLScheme: set your `appURLSceheme` here. Defaults to "spid-\(clientID)" if nil
      - parameter locale: Locale you want to use for requests - defaults to system
```
minarikg: Please add `bundle` to documentation as well

Source/Manager/ClientConfiguration.swift
```diff
@@ -189,7 +191,7 @@ public struct ClientConfiguration {
      - parameter appURLScheme: set your `appURLSceheme`. Defaults to "spid-\(clientID)" if nil
      - parameter locale: Locale you want IdentityManager to use for requests
```
minarikg: Same here

marcincs330:
We managed to remove instantiating `ClientConfiguration` from our test code, so I'm closing this one.

PR #260, opened 2019-05-27
Don't round the corners of the content view in IdentifierViewController.

zamzterz:
This shows up in the UI as grey background where the content view
doesn't cover it.

### DIFF
--- a/Source/UI/Screens/Identifier/IdentifierViewController.swift
+++ b/Source/UI/Screens/Identifier/IdentifierViewController.swift
@@ -27,7 +27,6 @@ class IdentifierViewController: IdentityUIViewController {
             self.backgroundView.backgroundColor = .schibstedLightGray
         }
     }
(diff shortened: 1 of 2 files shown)


Source/UI/Screens/Identifier/IdentifierViewController.swift
```diff
@@ -27,11 +27,7 @@ class IdentifierViewController: IdentityUIViewController {
             self.backgroundView.backgroundColor = .schibstedLightGray
         }
     }
-    @IBOutlet var contentView: UIView! {
-        didSet {
-            self.contentView.layer.cornerRadius = self.theme.geometry.contentGroupingCornerRadius
-        }
-    }
+    @IBOutlet var contentView: UIView!
```
aliak00: This is now unused so you can "disconnect" it from the view in the xib file and delete it.

PR #263, opened 2019-07-25
Enable Face ID

sa-mao:


Source/UI/Coordinators/PasswordCoordinator.swift
```diff
@@ -308,27 +308,35 @@ extension PasswordCoordinator {
...
+            context.canEvaluatePolicy(.deviceOwnerAuthenticationWithBiometrics, error: nil)
+            if context.biometryType == .touchID {
+                let viewModel = PasswordViewModel(
+                    identifier: identifier,
+                    loginFlowVariant: loginFlowVariant,
+                    localizationBundle: self.configuration.localizationBundle
+                )
+                let message = viewModel.biometricsOnboardingMessage
+                    .replacingOccurrences(of: "$0", with: configuration.appName)
+                let title = viewModel.biometricsOnboardingTitle
```
minarikg: Suggestion: Since the text in the title and message is dedicated to `TouchID`, you can use names: `touchIdOnboardingTitle` and `touchIdOnboardingMessage`

Source/UI/Coordinators/PasswordCoordinator.swift
```diff
@@ -308,27 +308,35 @@ extension PasswordCoordinator {
...
-            let viewModel = PasswordViewModel(
-                identifier: identifier,
-                loginFlowVariant: loginFlowVariant,
-                localizationBundle: self.configuration.localizationBundle
-            )
-            let message = viewModel.biometricsOnboardingMessage
-                .replacingOccurrences(of: "$0", with: configuration.appName)
-            let title = viewModel.biometricsOnboardingTitle
+            let context = LAContext()
+            context.canEvaluatePolicy(.deviceOwnerAuthenticationWithBiometrics, error: nil)
```
zamzterz: Why is this called? The return value isn't used for anything here.

minarikg: https://developer.apple.com/documentation/localauthentication/lacontext/2867583-biometrytype
> This property is set only after you call the canEvaluatePolicy(_:error:) method, and is set no matter what the call returns. The default value is LABiometryType.none.

minarikg: Although might be extracted to a `var` so it will be clear these 3 lines are together required to get `biometryType`'s value.
```
    private var biometryType: LABiometryType {
        let context = LAContext()
        // `biometryType` property is set only after you call the canEvaluatePolicy(_:error:) method, and is set no matter what the call returns. The default value is LABiometryType.none.
        context.canEvaluatePolicy(.deviceOwnerAuthenticationWithBiometrics, error: nil)
        return context.biometryType
    }
```

Source/UI/Screens/Password/PasswordViewModel.swift
```diff
@@ -88,10 +88,10 @@ extension PasswordViewModel {
     var biometricsPrompt: String {
         return "PasswordScreenString.biometricsPrompt".localized(from: self.localizationBundle)
     }
-    var biometricsOnboardingTitle: String {
+    var touchIdOnboardingTitle: String {
         return "PasswordScreenString.biometricsOnboarding.title".localized(from: self.localizationBundle)
```
minarikg: "PasswordScreenString.biometricsOnboarding.title" -> "PasswordScreenString.touchIdOnboarding.title"

Source/UI/Screens/Password/PasswordViewModel.swift
```diff
@@ -88,10 +88,10 @@ extension PasswordViewModel {
...
     }
-    var biometricsOnboardingMessage: String {
+    var touchIdOnboardingMessage: String {
         return "PasswordScreenString.biometricsOnboarding.message".localized(from: self.localizationBundle)
```
minarikg: same here `biometricsOnboarding`

minarikg:
Sample app's `Info.plist` should have:
```
    <key>NSFaceIDUsageDescription</key>
    <string>Use Face ID instead of a password to access your account.</string>
```

scripts/populate-dummy-secrets.sh
```diff
@@ -5,6 +5,8 @@ cat << 'EOF' > ./Example/Info.plist
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>NSFaceIDUsageDescription</key>
+	<string>Example will use Touch ID to improve your login experience</string>
```
zamzterz: Typo: "Touch ID" -> "Face ID"

zamzterz: Typo: "Touch ID" -> "Face ID"

Issue #272, opened 2019-08-01
Wrong behaviour for touch Id when you reinstall the app.


sa-mao:
Steps to reproduce:

1. Install the example app on your mobile, and set up touch id login.
2. Uninstall then reinstall the example app on your mobile.
3. Make sure that touch Id is enabled before you start your first login.
4. Start your first login.

The bug:
- When you start your first login, you are still prompt for your fingerprint. After which you are prompt for your consent to use touch Id.
- Regardless of your response, you will be logged in.

What it should do:
- Show the password view.
- After a successful login, ask for your consent to use touch Id.

PR #273, opened 2019-08-01
Fix: TouchID prompt on second install

sa-mao:
If merged this will fix this issue #272

### DIFF
--- a/Source/UI/Coordinators/PasswordCoordinator.swift
+++ b/Source/UI/Coordinators/PasswordCoordinator.swift
@@ -8,6 +8,7 @@ import UIKit
 
 private struct Constants {
     static let BiometricsSecretsLabel = "com.schibsted.account.biometrics.secrets"
+    static let HasLoggedInBeforeSettingsKey = "hasLoggedInBefore"
 }
 
 class PasswordCoordinator: AuthenticationCoordinator, RouteHandler {
@@ -241,6 +242,11 @@ extension PasswordCoordinator {
             // Fallback to passsword login
             return nil
         }
+
+        let hasLoggedInBefore = Settings.value(forKey: Constants.HasLoggedInBeforeSettingsKey) as? Bool
+        if hasLoggedInBefore == nil {
+            return nil
+        }
         var query = [String: Any]()
         query[kSecClass as String] = kSecClassGenericPassword
         query[kSecReturnData as String] = kCFBooleanTrue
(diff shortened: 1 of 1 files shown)


Source/UI/Coordinators/PasswordCoordinator.swift
```diff
@@ -8,6 +8,7 @@ import UIKit
 
 private struct Constants {
     static let BiometricsSecretsLabel = "com.schibsted.account.biometrics.secrets"
+    static let HasLoggedInBeforeSettingsKey = "hasLoggedInBefore"
```
minarikg: Maybe follow the pattern: "com.schibsted.account.hasLoggedInBefore"

Source/UI/Coordinators/PasswordCoordinator.swift
```diff
@@ -8,6 +8,7 @@ import UIKit
 
 private struct Constants {
```
minarikg: add: `private init() { }` to avoid possibility to create an instance of `Constants`

Source/UI/Coordinators/PasswordCoordinator.swift
```diff
@@ -241,6 +242,11 @@ extension PasswordCoordinator {
             // Fallback to passsword login
             return nil
         }
+
+        let hasLoggedInBefore = Settings.value(forKey: Constants.HasLoggedInBeforeSettingsKey) as? Bool
```
minarikg:
```
guard let hasLoggedInBefore = Settings.value(forKey: Constants.HasLoggedInBeforeSettingsKey) as? Bool else { return nil }
```

minarikg: You can also join with the previous guard
` guard #available(iOS 11.3, *) else {`

minarikg: Wouldn't this prevent users from reusing password after app upgrade?

Source/UI/Coordinators/PasswordCoordinator.swift
```diff
@@ -253,6 +258,7 @@ extension PasswordCoordinator {
         if status == noErr, let qresult = queryResult as? Data, let password = String(data: qresult as Data, encoding: .utf8) {
             return password
         } else {
+            Settings.setValue(false, forKey: Constants.HasLoggedInBeforeSettingsKey)
```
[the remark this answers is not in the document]
sa-mao: True, but I prefer to be explicit here.

zamzterz:
Since Apple keeps the data stored in the keychain even if the app is uninstalled, could we just keep using it without prompting for consent again on re-install?
@idanor

PR #278, opened 2019-09-13
Update to New schibsted logo

sa-mao:
Use the new schibsted Logo.


aliak00:
Vector images automatically scale cleanly. The equivalent would be to provide a number of PNG files scaled manually to your liking.

sa-mao:
Just updated the image to use an svg instead.

PR #284, opened 2019-10-23
Changed: remove "self" via SwiftFormat script

minarikg:
changed: scripts/format.sh
the rest is done by SwiftFormat

### DIFF
--- a/Source/Core/EventEmitter.swift
+++ b/Source/Core/EventEmitter.swift
@@ -19,13 +19,13 @@ class ReceiverHandle<Parameters>: Hashable {
 
     fileprivate init(object: AnyObject, handler: @escaping (AnyObject) -> (Parameters) -> Void) {
         self.object = object
-        self.objectHandler = handler
-        self.descriptionText = "AnyObject"
+        objectHandler = handler
+        descriptionText = "AnyObject"
     }
 
     fileprivate init(block: @escaping (Parameters) -> Void) {
         self.block = block
-        self.descriptionText = "block"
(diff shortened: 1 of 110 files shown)


Source/Core/EventEmitter.swift
```diff
@@ -19,13 +19,13 @@ class ReceiverHandle<Parameters>: Hashable {
 
     fileprivate init(object: AnyObject, handler: @escaping (AnyObject) -> (Parameters) -> Void) {
         self.object = object
```
aliak00: 😬

minarikg:
The reason for this to highlight the cases when self is retained by a closure. And think about potential retain cycles when you see “self”.
Xcode also does color highlights for properties vs local variables. Also functions should be simple and short enough to realize which are local variables.

Source/Core/EventEmitter.swift
minarikg: This is normal for an init in Swift.
or like ‘guard let self = self else {return}’ in closures.

aliak00:
> The reason for this PR is to highlight the cases when self is retained by a closure. And think about potential retain cycles when you see “self”.

Ah! That's a fair point. I'm not sure that's something people regularly think about when they see `self` but I see the advantage in that indeed.

>Xcode also does color highlights for properties vs local variables. Also functions should be simple and short enough to realize which are local variables.

Code changes hardly apply to an entire function. And while i agree with the theory whole heartedly, "simple and short" is sometimes just not practical sometimes :)

PR #288, opened 2019-11-08
Fix #280: Make login flow modal on iOS 13 if it isn't cancellable.

### LINKED ISSUE #280: modal presentation style of identity ui when compiled with iOS 13.x sdk
according to the documentation `modalPresentationStyle` behaviour changed:

> Defaults to UIModalPresentationAutomatic on iOS starting in iOS 13.0, and UIModalPresentationFullScreen on previous versions.
By default UIViewController resolves UIModalPresentationAutomatic to UIModalPresentationPageSheet, but other system-provided view controllers may resolve UIModalPresentationAutomatic to other concrete presentation styles.

because of this change when compiled with xcode 11 and iOS 13, IdentityUI is presented in a way that it makes it possible to dismiss by swiping down, even if `isCancelable` is set to false

IMG_0D5A3D4ABE4F-1
IMG_CE13BA07BC23-1

possible fixes include changing presentation style to `.fullScreen` or disabling swipe gesture when presented modally by setting `isModalInPresentation` to true
it might be good idea to tie this to `isCancelable` parameter of IdentityUI?

those changes also have implications on close/save/dismiss type of buttons on modal presentation of view controllers, here's great article that explains everything:
https://medium.com/@hacknicity/view-controller-presentation-changes-in-ios-13-ac8c901ebc4e

zamzterz:


zamzterz:
There's a [failing test](https://travis-ci.org/schibsted/account-sdk-ios/builds/609127534#L2858) that looks related to iOS 13 update:
```
Should_perform_a_standard_data_task, expected to equal <SchibstedAccountSDK/2.4.0
(iPhone; iOS 13.0)>, got <nil> (use beNil() to match nils)
```

Issue #294, opened 2020-02-20
Need to update the age copy to meet the recent changes in regulation.


sa-mao:
There has been a recent change in the consent age limit regulation in Norway. We need to update the age copy to meet that [change](https://www.datatilsynet.no/personvern-pa-ulike-omrader/kundehandtering-handel-og-medlemskap/digitale-tjenester-og-forbrukeres-personopplysninger/?id=12158).

PR #297, opened 2020-02-28
#294 Update age limit according to change in regulation.

zamzterz:


sa-mao:
Also need to change the date validation. 
https://github.com/schibsted/account-sdk-ios/blob/d2a0f8408056a7a3bcc5604ade8389d2ac258a14/Source/UI/Screens/RequiredFields/SupportedRequiredField.swift#L81

PR #300, opened 2020-03-29
Downscale PDF asset for Schibsted logo

vmagaziy:
Having the mentioned PDF asset in its original size (1698×355) may cause performance issues when the login dialog is interactively dismissed in iOS 13 because the asset is too large.

ezgif-6-e909bc084ad5

This PR decreases dimensions of the asset to 149 × 31 while aspect ratio is _maintained_:
```
1698 / 355 = 4.783
149 / 31 = 4.806
```


minarikg:
@vmagaziy 
- I’m wondering why is it rendereded multiple times during dismiss animation.
- Did you preserv the PDF as a vector image?

clausjoergensen:
> * Did you preserv the PDF as a vector image?

This is not a 3rd party view, it's the `IdentifierViewController.xib` using the asset from the `Images.xcassets`, and no, it is not being preserved as vector data. 

But the size is still absurd compared to the render size, using a smaller asset is better in every way possible. Smaller size footprint, less memory usage, smoother animations.

PR #309, opened 2020-10-09
Added: Support for Swift Package Manager

clausjoergensen:
In order to support Swift Package Manager the following changes had to be done:

## Changes

- `SPiDAccessToken` and `SPiDKeychainWrapper` was converted from Objective-C to Swift (as SwiftPM does not support mixed languages)
- `import Foundation` / `import UIKit` had to be added to all relevant files, now that there is no longer any Objective-C code to do global imports of those headers (which is a bad practice anyway)
- Swift generic types are implied non-escaping, so `@escaping` had to be added to the `object: T` parameter for the logging functionality.

## Problems

- I can't verify the `testTarget` because of how the project is setup (it requires a entitlement file, which you can't have for a unit test project).


salling:
@clausjoergensen I think it'd make sense to have the same deployment target in the Swift Package as in the Pod. That way, we don't get these warnings about deprecated APIs. iOS 9. WDYT?

PR #315, opened 2020-11-16
Fixed crash when NSKeyedUnarchiver.unarchiveObject tried to unpack ObjC version of the SPiDAccessToken

clausjoergensen:
After the type was converted to Swift, the full type name changed, which could cause a crash unarchiving the ObjC type due to a type name mismatch.


zamzterz:
Additional necessary fixes after manual testing added in #316

PR #319, opened 2020-11-17
Added support for logging in using shared web credentials

clausjoergensen:
Follow-up to #314 where we fetch the credentials from the shared web credentials, allowing the user to quickly login like seen here:

login


Source/UI/IdentityUI.swift
```diff
@@ -753,3 +774,45 @@ private final class DismissableNavigationController: UINavigationController {
...
 }
+
+private final class SharedWebCredentialsController: NSObject {
+    // Stored anonymously to retain the ASAuthorizationController
+    private var authorizationController: NSObject?
+
+    var didCompleteWithAuthorization: (String, String) -> Void = { _, _ in }
+    var didCompleteWithError: (Swift.Error) -> Void = { _ in }
+
+    override init() {
```
vmagaziy, line 796: Can we provide the result as a `Swift.Result` and add closure to the initialiser?

Source/UI/IdentityUI.swift
```diff
@@ -462,6 +468,21 @@ extension IdentityUI {
...
+                sharedWebCredentialsController.didCompleteWithError = { [weak self] _ in
+                    self?.show(input: .byLoginMethod(.password, presentingViewController: vc, localizedTeaserText: localizedTeaserText, scopes: scopes), client: client, completion: completion)
+                }
+                sharedWebCredentialsController.didCompleteWithAuthorization = { [weak self] user, password in
+                    guard let email = EmailAddress(user) else { return }
+                    self?._identityManager.login(username: .email(email), password: password, scopes: scopes, persistUser: true, useSharedWebCredentials: false, completion: { _ in
+                        completion(.skip)
+                    })
+                }
+                self.sharedWebCredentialsController = sharedWebCredentialsController
```
vmagaziy, line 493: Note: `authorizationController` has to be reset to avoid the retain cycle.

Source/UI/IdentityUI.swift
```diff
@@ -462,6 +468,21 @@ extension IdentityUI {
         case let .byRoute(route, vc):
             handleRouteForUnpresentedUI(route: route, byPresentingIn: vc, client: client, completion: completion)
         case let .byLoginMethod(loginMethod, vc, localizedTeaserText, scopes):
+            if #available(iOS 13.0, *), case .sharedWebCredentials = loginMethod {
+                let sharedWebCredentialsController = SharedWebCredentialsController()
+                sharedWebCredentialsController.didCompleteWithError = { [weak self] _ in
+                    self?.show(input: .byLoginMethod(.password, presentingViewController: vc, localizedTeaserText: localizedTeaserText, scopes: scopes), client: client, completion: completion)
+                }
+                sharedWebCredentialsController.didCompleteWithAuthorization = { [weak self] user, password in
+                    guard let email = EmailAddress(user) else { return }
```
vmagaziy: Here we have to fall back as well

Source/UI/IdentityUI.swift
```diff
@@ -136,6 +140,8 @@ public class IdentityUI {
     }
     private let _identityManager: IdentityManager
 
+    private var sharedWebCredentialsController: SharedWebCredentialsController?
```
vmagaziy, line 143: Can we avoid storing it? It should exist during fetching time only

Source/UI/IdentityUI.swift
clausjoergensen: What does "reset" mean in this context? You mean we have to explicitly set it to nil?

Source/UI/IdentityUI.swift
clausjoergensen: We need to retain it as the delegate is weak.

clausjoergensen: But I've changed it to be removed (nilled out) once the completion blocks are done.

Source/UI/IdentityUI.swift
```diff
@@ -753,3 +789,45 @@ private final class DismissableNavigationController: UINavigationController {
         fatalError("init(coder:) has not been implemented")
     }
 }
+
+private final class SharedWebCredentialsController: NSObject {
+    // Stored anonymously to retain the ASAuthorizationController
+    private var authorizationController: NSObject?
```
vmagaziy, line 795: Let's just do `let controller = ASAuthorizationController(authorizationRequests: [ASAuthorizationPasswordProvider().createRequest()])` and don't bother clearing it out as it will be gone along with `SharedWebCredentialsController`

Source/UI/IdentityUI.swift
```diff
@@ -462,6 +468,36 @@ extension IdentityUI {
         case let .byRoute(route, vc):
             handleRouteForUnpresentedUI(route: route, byPresentingIn: vc, client: client, completion: completion)
         case let .byLoginMethod(loginMethod, vc, localizedTeaserText, scopes):
+            if #available(iOS 13.0, *), case .sharedWebCredentials = loginMethod {
+                let fallback = { [weak self] in
+                    self?.show(input: .byLoginMethod(.password, presentingViewController: vc, localizedTeaserText: localizedTeaserText, scopes: scopes), client: client, completion: completion)
+                    self?.sharedWebCredentialsController = nil
+                }
+
+                sharedWebCredentialsController = SharedWebCredentialsController { [weak self] result in
```
vmagaziy, line 477: nit: we can have `sharedWebCredentialsController` captured by the closure, to it leaves until the closure it executed, but then we would have to provide a closure separately, not via the initialiser

Source/UI/IdentityUI.swift
clausjoergensen: We can't have a strong typed class member since it would require the entire class to be marked `@available(iOS 13.0, *)` which then cascades up. This is intentionally kept as anonymous type.

Source/UI/IdentityUI.swift
clausjoergensen: Yeah, but I dislike the idea of _intentionally_ using a closure to retain an object, rather than having a temporary class member.

Source/UI/IdentityUI.swift
```diff
@@ -23,6 +24,9 @@ public enum LoginMethod {
     /// asks for identifier and then a password to either login or signup if not already registered; the specified email will appear pre-filled in the identity
     /// UI, yet the user will still be able to modify it before submitting.
     case passwordWithPrefilledEmail(EmailAddress)
+    /// prompts the user for using their shared web credentials (if available)
+    @available(iOS 13.0, *)
+    case sharedWebCredentials
```
vmagaziy, line 29: Shouldn't these be called just `sharedCredentials` -- why are they `web`?

Source/UI/IdentityUI.swift
```diff
@@ -200,7 +206,7 @@ public class IdentityUI {
         let input = UpdatedTermsCoordinator.Input(currentUser: user, terms: terms)
 
         updatedTermsCoordinator = UpdatedTermsCoordinator(navigationController: navigationController, configuration: configuration)
-        updatedTermsCoordinator?.start(input: input) { _ in
+        updatedTermsCoordinator?.start(input: input) { output in
```
vmagaziy, line 209: nit: where the newly introduced `output` is used?

Source/UI/IdentityUI.swift
```diff
@@ -462,6 +468,11 @@ extension IdentityUI {
         case let .byRoute(route, vc):
             handleRouteForUnpresentedUI(route: route, byPresentingIn: vc, client: client, completion: completion)
         case let .byLoginMethod(loginMethod, vc, localizedTeaserText, scopes):
+            if #available(iOS 13.0, *), case .sharedWebCredentials = loginMethod {
```
vmagaziy, line 471: nit: Why do we have to additionally check available version as `sharedWebCredentials` is available as of iOS 13?

Source/UI/IdentityUI.swift
```diff
@@ -477,6 +488,105 @@ extension IdentityUI {
...
+                 completion: completion)
+
+            sharedWebCredentialsController = nil
+        }
+
+        sharedWebCredentialsController = SharedWebCredentialsController { [weak self] result in
+            guard case .success((let username, let password)) = result,
+                  let email = EmailAddress(username),
+                  let self = self else {
+                fallback()
```
vmagaziy, line 512: nit: Can we log the error here? It would be good to know whether it's just because of cancellation or invalid user name (malformed e-mail)

Source/UI/IdentityUI.swift
```diff
@@ -477,6 +488,105 @@ extension IdentityUI {
...
+                }
+
+                func cancel() {
+                    completion(.cancel)
+                }
+
+                func presentTerms() {
+                    self.identityManager.fetchTerms { result in
+                        guard case .success(let terms) = result else {
+                            cancel()
```
vmagaziy, line 539: Does it mean we will be just silently fail if we fail to fetch terms? Shall we logout in this case as well?

Source/UI/IdentityUI.swift
```diff
@@ -477,6 +488,105 @@ extension IdentityUI {
...
+
+                        updatedTermsCoordinator.start(input: .init(currentUser: user, terms: terms)) { output in
+                            IdentityUI.updatedTermsCoordinator = nil
+
+                            navigationController.dismiss(animated: true) {
+                                switch output {
+                                case .success:
+                                    success()
+                                case .cancel:
+                                    cancel()
```
vmagaziy, line 565: Same thing w/ logout

Source/UI/IdentityUI.swift
```diff
@@ -477,6 +488,105 @@ extension IdentityUI {
...
+                    }
+                }
+
+                user.agreements.status { result in
+                    guard case .success(let agreed) = result, agreed else {
+                        presentTerms()
+                        return
+                    }
+
+                    success()
```
vmagaziy, line 582: Why do we success if agreements cannot be checked?

Source/UI/Screens/Identifier/IdentifierViewController.swift
```diff
@@ -170,6 +170,8 @@ class IdentifierViewController: IdentityUIViewController {
             showPhoneNumber()
             countryCode.text = prefilledPhoneComponents.countryCode
             phoneNumber.text = prefilledPhoneComponents.number
+        case .sharedWebCredentials:
+            fatalError()
```
vmagaziy, line 174: What does this stand for?

Source/UI/IdentityUI.swift
```diff
@@ -477,6 +488,105 @@ extension IdentityUI {
...
+                return
+            }
+
+            self.identityManager.login(username: .email(email),
+                                       password: password,
+                                       scopes: scopes,
+                                       persistUser: true,
+                                       useSharedWebCredentials: false) { result in
+                guard case .success = result else {
+                    fatalError()
```
vmagaziy, line 522: Where is the guarantee login is always succeeding? What if user with stored e-mail is no longer valid or password got changed?

Source/UI/IdentityUI.swift
clausjoergensen: No, same naming convention as #314

Source/UI/IdentityUI.swift
clausjoergensen: Because you can't call a iOS 13 method without a #available check. It's required by the compiler.

Source/UI/IdentityUI.swift
clausjoergensen, line 209:
```suggestion
        updatedTermsCoordinator?.start(input: input) { _ in
```

Source/UI/IdentityUI.swift
clausjoergensen: We're not, you're reading the code wrong. We succeed if we read the agreements _and_ they were agreed to.

Source/UI/IdentityUI.swift
clausjoergensen: The logout is already called inside the `UpdatedTermsCoordinator`

Source/UI/IdentityUI.swift
clausjoergensen: The existing flow doesn't logout in this case, but it does present a popup with the error. Maybe we should do the same here.

Source/UI/IdentityUI.swift
clausjoergensen, line 522:
```suggestion
                    fallback()
```

clausjoergensen: This was a typo, meant to be `fallback()`

Source/UI/Screens/Identifier/IdentifierViewController.swift
clausjoergensen: This flow is impossible to hit (the view controller will never be presented with this login method)

vmagaziy:
Something is wrong with styling when T&C are presented in the flow:

Screen Shot 2020-11-18 at 10 13 26

clausjoergensen:
> Something is wrong with styling when T&C are presented in the flow:

Everything looks correct to me on that screenshot.

vmagaziy:
The properly styled T&C page looks like this:

Screen Shot 2020-11-18 at 10 19 47

Notice the colours of buttons and checkboxes.

clausjoergensen:
@vmagaziy the theme have been fixed now.

Source/UI/IdentityUI.swift
clausjoergensen: This now logs out the user and presents a popup with an error.

Source/Core/SynchronizedWeakDictionary.swift
```diff
@@ -18,19 +18,19 @@ class WeakValue<V: AnyObject>: CustomStringConvertible {
...
+            return weakValue.value
+        }
         set {
             guard newValue != nil else {
                 dictionary[key] = nil
                 return
             }
             dictionary[key] = WeakValue(newValue)
         }
-        get {
```
vmagaziy, line 28: Here

Source/Manager/IdentityManager.swift
```diff
@@ -370,7 +370,8 @@ public class IdentityManager: IdentityManagerProtocol {
      - parameter useSharedWebCredentials: whether the credentials should be saved to the shared web credentials.
      - parameter completion: a callback that is called after the credential is checked.
      */
-    public func login(username: Identifier, password: String, scopes: [String] = [], persistUser: Bool, useSharedWebCredentials: Bool, completion: @escaping NoValueCallback) {
+    public func login(username: Identifier, password: String, scopes: [String] = [], persistUser: Bool,
```
vmagaziy, line 373: Same

Source/Manager/IdentityManagerOverloads.swift
```diff
@@ -40,8 +40,10 @@ public extension IdentityManagerProtocol {
     }
 
     /// - SeeAlso: `IdentityManager.login(...)`
-    func login(email: EmailAddress, password: String, scopes: [String] = [], persistUser: Bool, useSharedWebCredentials: Bool, completion: @escaping NoValueCallback) {
-        return login(username: Identifier(email), password: password, scopes: scopes, persistUser: persistUser, useSharedWebCredentials: useSharedWebCredentials, completion: completion)
+    func login(email: EmailAddress, password: String, scopes: [String] = [], persistUser: Bool,
```
vmagaziy, line 43: Same

Source/Manager/LegacySPiDAccessToken/SPiDAccessToken.swift
```diff
@@ -50,7 +50,7 @@ open class SPiDAccessToken: NSObject, NSCoding {
         coder.encode(expiresAt, forKey: CodingKeys.expiresAt.rawValue)
         coder.encode(refreshToken, forKey: CodingKeys.refreshToken.rawValue)
     }
-
```
vmagaziy, line 53: Same

Source/Manager/Protocols/IdentityManagerProtocol.swift
```diff
@@ -26,7 +26,8 @@ public protocol IdentityManagerProtocol: AnyObject {
     func validate(oneTimeCode: String, for _: Identifier, scopes: [String], persistUser: Bool, completion: @escaping NoValueCallback)
 
     ///
-    func login(username: Identifier, password: String, scopes: [String], persistUser: Bool, useSharedWebCredentials: Bool, completion: @escaping NoValueCallback)
+    func login(username: Identifier, password: String, scopes: [String], persistUser: Bool,
```
vmagaziy, line 29: Same

Source/UI/Coordinators/PasswordCoordinator.swift
```diff
@@ -104,7 +104,13 @@ extension PasswordCoordinator {
         viewController.didRequestAction = { [weak self] action in
             switch action {
             case let .enter(password, shouldPersistUser):
-                self?.submit(password: password, for: identifier, on: loginFlowVariant, persistUser: shouldPersistUser, scopes: scopes, useSharedWebCredentials: useSharedWebCredentials, completion: completion)
+                self?.submit(password: password,
```
vmagaziy, line 107: Same

Source/UI/Coordinators/PasswordCoordinator.swift
```diff
@@ -148,7 +154,8 @@ extension PasswordCoordinator {
         }
 
         presentedViewController?.startLoading()
-        signinInteractor.login(username: identifier, password: password, scopes: scopes, useSharedWebCredentials: useSharedWebCredentials) { [weak self] result in
+        signinInteractor.login(username: identifier, password: password, scopes: scopes,
```
vmagaziy, line 157: Same

vmagaziy (commented):
@clausjoergensen -- can we reduce a number of changes, so we have only 2 files affected as before + required adjustments to tests?

clausjoergensen:
> @clausjoergensen -- can we reduce a number of changes, so we have only 2 files affected as before + required adjustments to tests?

No. We need the project to build to ensure full testing (and I'm sure you can review a few whitespace changes)

clausjoergensen:
I moved all the cleanup to #322

vmagaziy (approved):
lgtm

Source/UI/IdentityUI.swift
```diff
@@ -477,6 +491,115 @@ extension IdentityUI {
...
+                                            scopes: scopes,
+                                            persistUser: true,
+                                            useSharedWebCredentials: false) { result in
+                    switch result {
+                    case .failure(let error):
+                        log(from: self, error)
+                        fallback(.passwordWithPrefilledEmail(email))
+                    case .success:
+                        let user = self._identityManager.currentUser
+                        user.agreements.status { result in
```
zamzterz, line 584: Instead of this re-implementation of asking user to accept agreements, I think `CompleteProfileCoordinator` could be re-used.

Something like:
```swift
let coordinator = CompleteProfileCoordinator(navigationController: navigationController,
                                             identityManager: self._identityManager,
                                             configuration: self.configuration)
let profileCoordinator = UpdateProfileInteractor(currentUser: user,
                                                 loginFlowVariant: .signin,
                                                 tracker: self.configuration.tracker)
```

zamzterz (changes requested):
Nice one! 👍 

Unfortunately there's more complexity to handle in the case of successful login:
This correctly handles accepting terms, but there might also be required fields the user needs to enter data for. This is pretty much what `CompleteProfileCoordinator` captures. So I think that needs to be triggered/displayed in case of successful login with password from shared web credentials.

Source/UI/IdentityUI.swift
clausjoergensen: The problem is we can't present the other coordinators from this flow, as they don't have a completion block. This results in the consumer of the API (the app) doesn't know whether the login succeeded or not, and can't update it's UI.

clausjoergensen: And remember, we also want to do this without showing any UI unless necessary. From what I've seen, the coordinators flow _always_ presents something.

clausjoergensen: Ah, CompleteProfileCoordinator is special in that it checks all the paths and doesn't present anything unless it needs to. (That's different from the UpdateProfileCoordinator). That might work.

clausjoergensen:
@zamzterz I pushed a update using CompleteProfileCoordinator, I've verified it with the Terms&Conditions flow, seems to work.

clausjoergensen:
Hmm, it's not resetting correctly, so subsequent login attempts are failing identityUICompletion, looks like using the DismissableNavigationController doesn't work with the CompleteProfileCoordinator

clausjoergensen:
@zamzterz ah, no. Using this coordinator breaks the login flow because it attempts to present the UI. We can't have it present _anything_, no spinner, no nothing, unless it has to. 

This is what it looks like in a successful scenario for a user who don't need to agree / accept anything:

IMG_0591

zamzterz (approved):

Issue #189, opened 2018-08-24
 Make the checkbox for Remember Me optional 


mela124:
Summary
This is mainly an option for FINN. The reason we have this checkbox is to ensure that the user is aware of such behavior and a way for them to opt out of it. FINN will provide these information another way (not during the flow) so they would want a way to hide the checkbox. See their proposal here: https://docs.google.com/document/d/1g3sHVTQDBPu1NXn17bKjgtTEQQVcgBRwD2btsLokd5g/edit#heading=h.ouuojhde7gj6

- If it's hidden, then the user is considered still as opted-in as that's the default state of the checkbox
- By default, the checkbox should be shown
- 
-

3lvis:
I believe showing this checkbox on native apps is not a thing. Where did we get the idea to do this?

mela124:
Hi @3lvis, we discussed this with Thorild beforehand. It was needed to make sure the log in/sign up flow is GDPR-compliant. FINN suggested a different way to be GDPR-compliant but that is outside the flow that we provide thus we are thinking of making it optional. However, for those that won't do something similar as FINN, we still need to cover them therefore this functionality needs to be there.

We need to be able to be transparent to the user and provide them the opt out functionality that they will be stayed logged in on the app until they log out.  In the future, if we do autologin across apps, this even becomes more critical.

aliak00:
For this the config option should be added to `IdentityUIConfiguration` and persistence should be propagated from the `PasswordCoordinator` and the `PasswordlessCoordinator`

PR #323, opened 2020-11-20
New IdentityUIConfiguration option to be able to hide the persistent login checkbox

osanoj:
This is a fix for old issue https://github.com/schibsted/account-sdk-ios/issues/189. There is a good explanation for why we (FINN) need this in there. We have been using a fork for a long time, but I think it is better if we can get this changed and eventually stop using a forked version.


zamzterz:
Thanks for raising this (again)!
It needs more thorough discussion/user research on our end how to handle in the general case.

PR #330, opened 2021-08-26
1325-app transfer logic

Daniel-Echegaray:
Logic required for app transfer

### DIFF
--- a/Source/Manager/User/User+load.swift
+++ b/Source/Manager/User/User+load.swift
@@ -24,10 +24,10 @@ public extension User {
     /**
      Enables App transfer step 2
      */
-    static func loadFromDevice(withConfiguration clientConfiguration: ClientConfiguration, storageKey: String) -> User {
+    static func loadFromDevice(withConfiguration clientConfiguration: ClientConfiguration, storageKey: String) -> User? {
         let user = User(clientConfiguration: clientConfiguration)
         try? user.loadFromDeviceToKeychain(key: storageKey)
 
-        return User.loadLast(withConfiguration: clientConfiguration)
+        return user
     }
 }
--- a/Source/Manager/User/User+load.swift
+++ b/Source/Manager/User/User+load.swift
@@ -17,16 +17,16 @@ public extension User {
     /**
      Enables App transfer step 1
      */
-    static func storeOnDevice(withConfiguration clientConfiguration: ClientConfiguration, storageKey: String) {
+    static func storeOnDevice(withConfiguration clientConfiguration: ClientConfiguration, storageKey: String) throws {
         let user = User(clientConfiguration: clientConfiguration)
-        try? user.storeOnDevice(key: storageKey)
+        try user.storeOnDevice(key: storageKey)
     }
     /**
      Enables App transfer step 2
      */
-    static func loadFromDevice(withConfiguration clientConfiguration: ClientConfiguration, storageKey: String) -> User? {
+    static func loadFromDevice(withConfiguration clientConfiguration: ClientConfiguration, storageKey: String) throws -> User {
         let user = User(clientConfiguration: clientConfiguration)
-        try? user.loadFromDeviceToKeychain(key: storageKey)
+        try user.loadFromDeviceToKeychain(key: storageKey)
 
         return user
     }
--- a/Source/Manager/User/User+load.swift
+++ b/Source/Manager/User/User+load.swift
@@ -14,4 +14,20 @@ public extension User {
         try? user.loadStoredTokens()
         return user
     }
+    /**
+     Enables App transfer step 1
+     */
+    static func storeOnDevice(withConfiguration clientConfiguration: ClientConfiguration, storageKey: String) {
+        let user = User(clientConfiguration: clientConfiguration)
+        try? user.storeOnDevice(key: storageKey)
+    }
+    /**
+     Enables App transfer step 2
+     */
+    static func loadFromDevice(withConfiguration clientConfiguration: ClientConfiguration, storageKey: String) -> User {
+        let user = User(clientConfiguration: clientConfiguration)
+        try? user.loadFromDeviceToKeychain(key: storageKey)
+
+        return User.loadLast(withConfiguration: clientConfiguration)
+    }
 }
--- a/Source/Manager/User/User.swift
+++ b/Source/Manager/User/User.swift
@@ -292,13 +292,15 @@ public class User: UserProtocol {
         isPersistent = true
     }
 
+    
+    private static let keyPrefix = "old-sdk-app-transfer-"
+    
     func storeOnDevice(key: String) throws {
         do {
             let tokens = try UserTokensStorage(accessGroup: clientConfiguration.accessGroup).loadTokens()
-            let keyPrefix = "old-sdk-app-transfer-"
-
+            
             if let encoded = try? JSONEncoder().encode(tokens) {
-                UserDefaults.standard.set(encoded, forKey: keyPrefix + key)
+                UserDefaults.standard.set(encoded, forKey: User.keyPrefix + key)
             }
         } catch {
             log(level: .error, from: self, "failed to storeOnDevice tokens: \(error)", force: true)
@@ -306,9 +308,8 @@ public class User: UserProtocol {
     }
 
     func loadFromDeviceToKeychain(key: String) throws {
-        let keyPrefix = "old-sdk-app-transfer-"
 
-        if let tokenData = UserDefaults.standard.object(forKey: keyPrefix + key) as? Data,
+        if let tokenData = UserDefaults.standard.object(forKey: User.keyPrefix + key) as? Data,
            let decodedTokenData = try? JSONDecoder().decode(TokenData.self, from: tokenData) {
             // Storing in Keyhcain
             do {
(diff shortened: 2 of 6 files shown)


Source/Manager/User/User+load.swift
```diff
@@ -14,4 +14,20 @@ public extension User {
...
+        try? user.storeOnDevice(key: storageKey)
+    }
+    /**
+     Enables App transfer step 2
+     */
+    static func loadFromDevice(withConfiguration clientConfiguration: ClientConfiguration, storageKey: String) -> User {
+        let user = User(clientConfiguration: clientConfiguration)
+        try? user.loadFromDeviceToKeychain(key: storageKey)
+
+        return User.loadLast(withConfiguration: clientConfiguration)
```
mohsen-schibsted, line 31: We are creating an instance here, load something on it and then return from the object itself and not from the instance? 
Then why do we need the instance? 
Are we doing some static thing inside `loadFromDeviceToKeychain`

Source/Manager/User/User.swift
```diff
@@ -292,6 +292,38 @@ public class User: UserProtocol {
...
+            }
+        } catch {
+            log(level: .error, from: self, "failed to storeOnDevice tokens: \(error)", force: true)
+        }
+    }
+
+    func loadFromDeviceToKeychain(key: String) throws {
+        let keyPrefix = "old-sdk-app-transfer-"
+
+        if let tokenData = UserDefaults.standard.object(forKey: "old-sdk-app-transfer-" + key) as? Data,
```
mohsen-schibsted, line 311: you have this `"old-sdk-app-transfer-"` stored in `keyPrefix` but you are not using it

Source/Manager/User/User+load.swift
mohsen-schibsted: aha I think I see it now👍

`return User.loadLast(withConfiguration: clientConfiguration)` is the new way, but we want to remove the old way too and move the stored data on UserDefautls to Keychain?

Source/Manager/User/User.swift
markusbergh: Is it not user on line 219?

mohsen-schibsted: 320? Yes, but it can be used on line 311 too. So we prevent a typo bug.

Daniel-Echegaray: Thats better. I added it as static 🪚

Source/Manager/User/User+load.swift
Daniel-Echegaray: Not sure what you mean. But the App transfer will constitue of two steps.

Step 1 (Code line 20): Storing keychain data to UserDefaults.
Step 2 (27): Storing UserDefaults data to keychain.

The code line: _return User.loadLast(withConfiguration: clientConfiguration)_
Perhaps could be omitted, and instead having the adopting sdk consumer call it when they want to.

mohsen-schibsted: I am very new here so I have no info about the underlying code😁 let me read some more.

Now the question is why do we have to do it in two steps?🤔 why not just have both in one step? We are using UserDefault for something that is very sensitive. Probably you are gonna save it there for just a second, right? But many things can go wrong.

https://developer.apple.com/forums/thread/97231

Daniel-Echegaray: > I am very new here so I have no info about the underlying code😁 let me read some more.

Ah no worries. :D

> Now the question is why do we have to do it in two steps?🤔 why not just have both in one step? We are using UserDefault for something that is very sensitive. Probably you are gonna save it there for just a second, right? But many things can go wrong.
> 
> https://developer.apple.com/forums/thread/97231

There is a risk yes, although I think it is outweighed by the value of having users not being logged out, on doing an App transfer 🤔 . As the keychain gets cleaned out when doing an App transfer https://developer.apple.com/library/archive/technotes/tn2311/_index.html#//apple_ref/doc/uid/DTS40014135-CH1-A_ONE_TIME_LOSS_IN_KEYCHAIN_DATA_WILL_OCCUR_IF_YOU_SWITCH_YOUR_APP_ID_PREFIX

lindanordstrom: Why do we need to call `User.loadLast(withConfiguration: clientConfiguration)` from here, isn't it enough to return `user` ?

Source/Manager/User/User.swift
Daniel-Echegaray: resolving this conversation

Source/Manager/User/User+load.swift
Daniel-Echegaray: Thats right! Calling `User.loadLast(withConfiguration: clientConfiguration)`  is an unnecessary extra step. Omitting

Source/Manager/User/User.swift
```diff
@@ -292,6 +292,38 @@ public class User: UserProtocol {
...
     }
 
+    private static let keyPrefix = "old-sdk-app-transfer-"
+
+    func storeOnDevice(key: String) throws {
+        do {
+            let tokens = try UserTokensStorage(accessGroup: clientConfiguration.accessGroup).loadTokens()
+
+            if let encoded = try? JSONEncoder().encode(tokens) {
+                UserDefaults.standard.set(encoded, forKey: User.keyPrefix + key)
```
clausjoergensen, line 302: `UserDefaults` seems like a poor choice for this. https://cocoacasts.com/ud-11-why-do-you-lose-data-stored-in-user-defaults

At minimum you'd need to flush immediately after setting this.

Source/Manager/User/User+load.swift
```diff
@@ -14,4 +14,20 @@ public extension User {
         try? user.loadStoredTokens()
         return user
     }
+    /**
+     Enables App transfer step 1
+     */
+    static func storeOnDevice(withConfiguration clientConfiguration: ClientConfiguration, storageKey: String) {
+        let user = User(clientConfiguration: clientConfiguration)
+        try? user.storeOnDevice(key: storageKey)
```
clausjoergensen, line 22: Hiding the error here seems like a poor choice. The apps won't know it failed, and won't retry to store the key again.

Source/Manager/User/User+load.swift
```diff
@@ -14,4 +14,20 @@ public extension User {
...
+    }
+    /**
+     Enables App transfer step 2
+     */
+    static func loadFromDevice(withConfiguration clientConfiguration: ClientConfiguration, storageKey: String) -> User? {
+        let user = User(clientConfiguration: clientConfiguration)
+        try? user.loadFromDeviceToKeychain(key: storageKey)
```
clausjoergensen, line 29: (Same issue with hiding the error here)

Source/Manager/User/User.swift
```diff
@@ -292,6 +292,38 @@ public class User: UserProtocol {
         isPersistent = true
     }
 
+    private static let keyPrefix = "old-sdk-app-transfer-"
+
+    func storeOnDevice(key: String) throws {
```
clausjoergensen, line 297: Should use `rethrows`

Source/Manager/User/User.swift
Daniel-Echegaray: Interesting 🤓

Daniel-Echegaray: Although article says: 

> "Data loss isn't something you need to worry about if you're working with the defaults system. The UserDefaults class ensures any changes are written to disk at the appropriate time without impacting the performance of your application. It goes without saying that unexpected terminations of your application break this behavior, but that applies to any persistence solution."

Daniel-Echegaray: i'll add an invocation to [synchronise](https://developer.apple.com/documentation/foundation/userdefaults/1414005-synchronize)

clausjoergensen: Yes, synchronise

Source/Manager/User/User+load.swift
Daniel-Echegaray: Sure. I'll propagate up the error to the apps 👌

Source/Manager/User/User+load.swift
Daniel-Echegaray: Sure. I'll propagate up the error

Source/Manager/User/User.swift
Daniel-Echegaray: Why rethrows though?
_"A function or method can be declared with the rethrows keyword to indicate that it throws an error only if one of its function parameters throws an error"_
https://docs.swift.org/swift-book/ReferenceManual/Declarations.html

Source/Manager/User/User.swift
```diff
@@ -292,6 +292,38 @@ public class User: UserProtocol {
...
+
+            if let encoded = try? JSONEncoder().encode(tokens) {
+                UserDefaults.standard.set(encoded, forKey: User.keyPrefix + key)
+            }
+        } catch {
+            log(level: .error, from: self, "failed to storeOnDevice tokens: \(error)", force: true)
+        }
+    }
+
+    func loadFromDeviceToKeychain(key: String) throws {
```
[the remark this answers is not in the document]
Daniel-Echegaray: There is no throwing funciton parameter

clausjoergensen (approved):

lindanordstrom (approved):

PR #332, opened 2021-09-16
Adding: Clear stored user on device

minarikg:


Source/Manager/User/User+load.swift
```diff
@@ -21,6 +21,15 @@ public extension User {
         let user = User(clientConfiguration: clientConfiguration)
         try user.storeOnDevice(key: storageKey)
     }
+
+    /**
+     Enables App transfer step 1 - Clear when user logs out
```
Daniel-Echegaray, line 26: Comment is wrong. Does not "Enables App transfer step 1"

Daniel-Echegaray (approved):
lgtm! See comment

